PR #57, opened 2018-05-24
Webfilter

Hantse:
Hello,

I have create the filter attribute.

For the nuget your must add this depedency : Microsoft.AspNetCore.Mvc.Core

Use is simple, just add  [SyncResult] on controller or specific method. By default, all methods is trig (*), but you can define methods by : 
[SyncResult(OnMethods = new string[] { "POST" })]

Or simply with constructor :
[SyncResult("POST")]
[SyncResult("POST,DELETE")]


Mimetis:
Thx for your pull request @Hantse.
This PR is not included yet in the last version, but will be in the next version !

PR #69, opened 2018-08-02
AspNet Web API 2 support

gentledepp:
Hello!
I modified the project "Dotmim.Sync.Web" so that it creates dlls for
*) netstandard2.0 and ASP.NET Core (as up until now)
*) net462 and Web API 2
I followed this approach: [Replace existing projects with a multi-targeted .NET Core project](https://docs.microsoft.com/en-us/dotnet/core/porting/project-structure#replace-existing-projects-with-a-multi-targeted-net-core-project)
along with this [Sample](https://github.com/dotnet/samples/tree/master/framework/libraries/migrate-library-csproj/)

In order to create a nuget package, just use the "pack" command on the project as usual.

**Please do not merge yet - I am still testing.**

Thinks to keep in mind:
*) Deliberately dropped support for the SyncResultAttribute for Web APi 2. Reason: It could be created by deriving from ActionFilterAttribute, but DI support is a problem. It would require users to always use DI and, moreover, use a DI framework which supports property injection, as ctor injection is not possible with attributes in Web API 2.


Mimetis:
hey @gentledepp 
Thx for all your commits, it's so amazing !! Thx for all your great job and contribution to `Dotmim.Sync`

I don't have checked everything, but it seems everything is ok.

Anyway, I have some questions. My goal is to have the less projects as possible.   
Regarding `.Net Standard 2.0` and this tweet from @terrajobst ( https://twitter.com/terrajobst/status/1031999730320986112 ) 

We should target `.Net FX 4.7.2` to be able to use `.Net Standard 2.0`

So far, my questions:

- If we are targetting `.Net Fx 4.7.2 +` and `.Net Core 2.0+`, can we have only one project ?
- Do you need `.Net Fx  4.6` ?

gentledepp:
Hi!
Currently, our software platform runs with net462.
But as I see that Windows Server 2008 R2 is still supported by 472, I see no problems switching to that.

1) As for "as few projects as possible":
We only added one sample- and one test project for net462 as there seem to be issues with test runners when it comes to multi platform test projects.
For the *real* Dotmim.Sync support for Web Api 2, no project was added.
We only modified the Dotmim.Sync.Web.Server project to also target net462  => so you can also just package this project and get a single nuget for both, netstandard and net462

To see an explanation how this works, I provided a link in the first PR message (scroll to the top ;-))

2) I see unit tests are sometimes failing and sometimes not - both, on my machine and on appveyor. There seem to be side effects between test runs. I expect that your usage of IClassFixture<> may be the culprit, as you create a database once for all tests, but do not clean them up (truncate the tables, etc)
Could you please help me investigate that?

Mimetis:
1) I'm currently reviewing the code. Sorry for my miss understanding of your PR. As you said, there is no additionnal projects, so IT'S NICE !!! :)

2) I saw this weird bugs and some circumstances, I did not find the reason for now. I will continue to search. ANY suggestions will be really appreciated !! :)

gentledepp:
2.  Well... it is always the integration tests - at least as far as I can see.
What I am currently trying is to make sure, that each tests works also if executed separately. You seem to assume that the whole fixture is run (thus the TestPriority(1-12) stuff?)
But: When any test fails, or is executed in parallel (does xUnit do that?), the fact that those tests use one single database could tell us why there are side effects that cause those tests to fail occasionally

gentledepp:
1. yeah... it is... I really love the new project format. Well done Microsoft ;-)

Mimetis:
The problem I see is that we will have 2 binaries (for `Dotmim.Sync.Web.Server`) One for .Net 4.6 and one for .Net Core 2.0.   
So we will have eventually 2 binaries to deploy on nuget ?

gentledepp:
Yes.
There will be one nuget package, containing a subfolder for .net core and net472.
Nuget will automatically choose the right dll at build time. (this is kind of what we did back in the days with PCL - called bait and switch - where the PCL library contains only the interface or an abstract class which throws notimplementedexceptions, and for each target platform (android, ios, net462) there is a separate dll with the same name, but a platform specific implementation)

Mimetis:
Oh yes, make sense. I lost my brain during those hollidays :D

so far, until we are comitted to .net 4.7.2+ I guess we can have 2 nugets packages !

Will merge this PR before end of day !

gentledepp:
Just right-click dotmim.sync.web.server and select "pack"
then investigate the resulting nuget package: [screenshot](https://1drv.ms/u/s!AhVbOiFFuzx_2TnEVMBTqy-ClnPP)

It is 1 nuget for both platforms (asp.net core and asp.net (legacy)), both also having different nuget dependencies.

So the goal is to make the usage of dotmim.sync.web.server completely transparent to asp.net core and asp.net developers.

The only thing that bothers with the current package, is that it targets .net462 and .netstandard2.0, where it should basically target .net core instead of .net standard, as it has dependencies to asp.net core.... or am I wrong in this assumption? (I did not have time to play around with .net core yet)

gentledepp:
please do not merge yet! I want to make sure that the UTs pass!
...at least all except of the two that also fail on the master branch - maybe you can have a look at them ;-)

- MySqlAllColumnsTests.OneRowFromServer
- SyncConflictsTests.ConflictOnUniqueConstraint

gentledepp:
@Mimetis 
If you like you can add me via skype (same name as here) - maybe that makes it easier to discuss things. I will contribute quite a lot the upcoming weeks, as we *need* this framework now and some features (specifying columns to sync, adding a filter that does not exist on the table and applying pre- and postfixes to mssql triggers) are really needed by us.

Issue #63, opened 2018-07-18
Could Not Load Assembly 'System.Threading.Tasks.Extension, Version 4.2.0.0


josephorjinta:
Hi @Mimetis ,
I am trying to implement a ASP.NET Core 2.0 Web Proxy for Sync between SQL Server Database  on Server and SQLite Db meant to be persisted on a mobile device using Cross Platform (Xamarin Forms )Template,  Visual Studio 2017 Community Version 15.6.4. I have created the Web Proxy App using ASP.NET Core Web App (2.0), but when i  to implement the Sync function on .NET Standard Shared Code Project by typing

`public static async Task Sync`
`{`
 ` SqliteSyncProvider clientProvider = new SqliteSyncProvider("GasMobile.db");`
  ` var proxyClientProvider = new Dotmim.Sync.Web.WebProxyClientProvider(new Uri("http://localhost:54593/api/values"));`

    ` var agent = new SyncAgent(clientProvider, proxyClientProvider);`

     `var s = await agent.SynchronizeAsync();`
`}`

I had to add  Dotmim.Sync.Web NuGet Package to the SolidChain .Net Standard Code Sharing Project , On running the SolidChainGas.Android project on Visual Studio, It came up with the following error:

	Exception while loading assemblies: System.IO.FileNotFoundException: Could not load assembly 'System.Threading.Tasks.Extensions, Version=4.2.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. Perhaps it doesn't exist in the Mono for Android profile?
File name: 'System.Threading.Tasks.Extensions.dll'
   at Java.Interop.Tools.Cecil.DirectoryAssemblyResolver.Resolve(AssemblyNameReference reference, ReaderParameters parameters) in /Users/builder/data/lanes/5809/a7829590/source/monodroid/external/xamarin-android/external/Java.Interop/src/Java.Interop.Tools.Cecil/Java.Interop.Tools.Cecil/DirectoryAssemblyResolver.cs:line 229
   at Xamarin.Android.Tasks.ResolveAssemblies.AddAssemblyReferences(DirectoryAssemblyResolver resolver, ICollection`1 assemblies, AssemblyDefinition assembly, Boolean topLevel)
   at Xamarin.Android.Tasks.ResolveAssemblies.AddAssemblyReferences(DirectoryAssemblyResolver resolver, ICollection`1 assemblies, AssemblyDefinition assembly, Boolean topLevel)
[...]

PR #75, opened 2018-08-08
Web client/server separately

ashalkhakov:
I've made some changes here, mainly for Linux compatibility. Initially I started working on this because we hit #63, and I thought that it was due to the `Web` assembly being too big. :-)

* if tests are run on non-Windows, then SQL Server for Linux can be used instead of LocalDB (previously tests simply failed)
  * to create this container in Docker: `docker run -e 'ACCEPT_EULA=Y' -e 'SA_PASSWORD=QWE123qwe' -e 'MSSQL_PID=Express' -p 1433:1433 -d microsoft/mssql-server-linux:latest`
* again on Linux, it's easy to test against MySQL if container is created in docker like this:
    * `docker run --name mysql -p 3306 -e 'MYSQL_ROOT_PASSWORD=azerty31$' -d mysql:latest --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci --lower-case-table-names=1`
    * we set `lower_case_table_names`, otherwise `MySqlSyncForeignKeysTests` fail due to case-sensitivity of identifiers, which are lower-cased by `Dotmim.Sync`


ashalkhakov:
Next step: add a Xamarin-based Android app.

Mimetis:
Why split in two projects ? (client / server)
I mean, I didn't check everything for now (hollidayyyyyyyyyys) but it seems that both projects have the exact same references on asp.net core / mvc / session

So why the split ?

Sebastien

ashalkhakov:
You are correct, I forgot to remove them! Could you take a look now?

ashalkhakov:
> I try to keep the less projects as possible. Is it possible to have only two projects : Web.Client and Web.Server ?

Sure, let's trim it down.

> For now I have a problem with the Microsoft.Net.Http.Headers reference.

Yeah, we just hit this on a Windows machine. I'm looking into this.

ashalkhakov:
Thanks! Now we can move forward with a Xamarin test app!

Issue #76, opened 2018-08-10
CI pipeline


ashalkhakov:
Let's add a CI pipeline to this project.

The problem is I don't know which can we use, since this is an open source project. How about AppVeyor, they have a free plan for OSS projects. Or CircleCI, TravisCI or whatever else there is?

gentledepp:
Do you have any experiences with one of those?
What's important is to have a running instance of MSSQL and MySQL, as lots of the tests are integration tests.
My lucky guess would be using some preconfigured docker image, but I have no experience with that. Maybe you can make a PR? :-)

ashalkhakov:
> Maybe you can make a PR? :-)

I will work on this sometime next week. I don't have experience with any of these so I don't know if they allow to use databases during test runs.

gentledepp:
Great!
I have seem some successful projects (e.g. AutoMapper) using appveyor.
A short google search shows, that they do support MySQL and several MSSQL versions:

https://www.appveyor.com/docs/services-databases/

Maybe that is the best place to start :-)

Mimetis:
Im from Microsoft so we can use Visual Studio Team Services if you want ?

gentledepp:
sure - if the setup is equally simple 👍 
But appveyor is free for open source projects - as far as I know

ashalkhakov:
Please take a look. I will probably reopen another PR with cleaner git history. :)

ashalkhakov:
Now that #79 has been merged, I guess the only thing missing is probably nuget packaging and publishing. And there is also #71, about versioning (I guess we should adopt semantic versioning, but how does that blend with the proposal?).

@Mimetis what do you think? I have little idea how others are doing it. @gentledepp do you have any ideas regarding the general workflow? IIUC, we want "stable" builds to produce NuGet packages, which are then published.

gentledepp:
Basically I would suggest:
1) keep the assembly version at 1.0.0.0 and only raise the assembly file version
2) use semantic versioning for assembly file version and nuget package version
=> i.e. only raise first version part in case there are any breaking changes

Basically, I am a very strong supporter of backwardly compatibility, so any changes to a library that need modifications by the client projects should be avoided.

As for packaging - I guess this should not happen in CI directly.
There should be a process to release a new version which can be kicked-off manually - whenever the owners of the repository think it is time for it and not on every CI build (PR, daily build, whatever)

[...]

PR #78, opened 2018-08-13
CI

ashalkhakov:
This adds some building and testing for #76.

Things that DON'T work yet:

* SQL Server & MySQL tests fail due to connection strings (easily fixable, I guess)
* SQL Server on Linux will probably also fail (fixed in #75, where database provisioning does not work)
* packaging and publishing to NuGet (this requires separate steps)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,6 +32,4 @@ build_script:
 - dotnet build Projects/dotnet-sync/dotnet-sync.csproj
 
 test_script:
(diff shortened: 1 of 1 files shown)


gentledepp:
Thank you so much for your contribution!
I took your yml as foundation and created another PR which now also runs the integration tests. (only 2 fail due som dotmim.sync bugs)
PR #79

ashalkhakov:
@gentledepp thanks for the follow-up! I'm closing this in favor of #79.

Issue #83, opened 2018-08-23
[Issue] integration tests fail every now and then


gentledepp:
It seems the integration tests do have some side effects which cause them to fail every now and then.

I will investigate and come back with a PR

gentledepp:
Fixed by PR #84

PR #84, opened 2018-08-23
83 ci fixes

gentledepp:
Added assembly attribute 

      [assembly: CollectionBehavior(DisableTestParallelization = true)]

=> tests are now executed sequentially (not parallel) which means they are slower, but they do not fail

Fixed trigger creation by re-adding the standard postfix "_trigger" to SQLite,MSSQL,MySQL.
If no custom prefix was provided the triggers names would clash with those of the stored procedures. This caused nearly all tests to fail.

This fixed issue #83

### DIFF
--- a/Projects/Dotmim.Sync.Core/Serialization/DmBinaryConverter/Converters/ObjectConverter.cs
+++ b/Projects/Dotmim.Sync.Core/Serialization/DmBinaryConverter/Converters/ObjectConverter.cs
@@ -1,21 +1,20 @@
 ﻿using System;
-using System.Collections.Generic;
+using System.Collections.Concurrent;
 using System.Globalization;
-using System.Text;
 
 namespace Dotmim.Sync.Serialization.Converters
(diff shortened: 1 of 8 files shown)


gentledepp:
general findings on the topic "integration test side effects"
1) tried to fix by disposing all objects (SyncAgent, etc.) in fixures - no success
2) added assembly attribute to un-paralellize test execution => fixes the problem
3) searched for static fields in dotmim.sync that could cause these issues. 
Found 2 static non thread-safe dictionaries (ObjectConverter and SqlSyncAdapter) => changed them to concurrentdictionaries (but still no success)
4) tried to change the class-fixtures (where database is created and dropped after *all* tests have been run) to normal fixture (meaning: database was created and dropped for each single unit test) - did not fix the problem.

Mimetis:
The only thing I've notived so far, is that those weird bugs occurs when we are using BinarySerializer.

Issue #81, opened 2018-08-16
Filter - only one sync filter per table supported?


gentledepp:
Hi!

In my sync scenario, I need to pass two filters to the select_changes stored procedure:
one denotes the tenant (tenantid), and one the current user (userId)

I noticed, that when I add the second column (userId) as filter, dotmim.sync does not create any tracking tables, procedures or whatever and just fails silently.

Now initially I was thinking that the problem is, that the second colum (userid) actually does not exist on the table, that it filters.

However, when I added the *userid* column to the table, just to try it out, still no tracking tables, procedures, ... were created by dotmim.sync

So what I guess is, that currently only one filter value is supported.
Is that correct?

Because if so, this is a major show stopper for us => we'd need to implement it.

Mimetis:
Thx for feedback.
I never tried to make a multi filters scenario.
Let me think about it, and I will go back to you once I have made some tests

gentledepp:
I am taking over now.
What I do not understand, is why you create two distinct selectChanges stored procedures

- one with the default name
- one with the filter name included and also the filter parameter

Whenever you *do* want to apply a filter in the sync process, you must always sync with that filter. Otherwise you will get all the other rows that you don't like and they stay on your client device, even if you start to use the filters again.
So honestly: I do not get why you do that.

I was expecting there was always only one selectChanges stored procedure - either with or without filter(s) (just as specified by the application developer)

gentledepp:
@Mimetis 
I am currently implementing this. PR will follow shortly.
Could you please elaborate on the following details:

1) why are there multiple selectchanges stored procedures (filtered and unfiltered) - I guess you want to let the apps decide whether to filter or not?
2) Could you explain, how exactly the select changes stored procedure works?
  - Why are the filters also maintained in the tracking tables?
  - I created some test "SyncFilterMultipleFilters" where I copy your original SyncFilterTests and add an additional filter (StatusValue) into the mix. The stored procedure I come up with then looks like the following:

        CREATE PROCEDURE [dbo].[ServiceTickets_CustomerIDStatusValue__selectchanges]
        	@sync_min_timestamp bigint,
        	@sync_scope_id uniqueidentifier,
        	@sync_scope_is_new bit,
        	@sync_scope_is_reinit bit,
        	@CustomerID int,
        	@StatusValue int
        AS
        BEGIN
        SELECT 	[side].[ServiceTicketID], 
        	[base].[Title], 
        	[base].[Description], 
[...]

PR #85, opened 2018-08-23
81 virtual filters

gentledepp:
Adds the notion of virtual filters.
Fixes issue #81

### DIFF
--- a/Projects/Dotmim.Sync.Core/Builders/DbBuilder.cs
+++ b/Projects/Dotmim.Sync.Core/Builders/DbBuilder.cs
@@ -107,7 +107,7 @@ namespace Dotmim.Sync.Builders
             try
             {
                 var trackingTableBuilder = CreateTrackingTableBuilder(connection, transaction);
-                trackingTableBuilder.Filters = this.FilterColumns;
+                trackingTableBuilder.Filters = this.FilterColumns.GetColumnFilters();
 
                 if (trackingTableBuilder.NeedToCreateTrackingTable())
                 {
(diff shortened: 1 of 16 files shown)


Mimetis:
Thx @gentledepp .

Is it working on MySQL as well ? (did not see any MySQL files changed in the PR)

Seb

gentledepp:
No, not I wanted to hear your opinion veford I do a complete implementation.

gentledepp:
I do not quite understand this. SqlDbType and DbType are used everywhere throughout Dotmim.Sync.Core and are part of netstandard. What am I missing here?

Are you planning to support other sync providers as well, which are not SQL-based? (couchdb, ravendb, filesystem, ...)?

gentledepp:
Ok.. so I think I got your point.
I switched to DbType since it is also used in DmTable.
Is that ok with you? If 'yes', I'd provide a MySql implementation as well

gentledepp:
@Mimetis 
As you probably remember, we are working on a better filter aproach (which i outlined a bit here: #2 )

Now we managed to put it to work.
You can actually have a look at a working Dotmim.Sync-based implementation in the following repository: https://github.com/gentledepp/SyncFilters

What it basically does is solve the "getting out of scope" problem of the old sync framework, so you can define very fine grained which user gets to see which rows - also based on permissions which may be totally independent from the base row that actually gets synchronized.

Mimetis:
Ok let's go ! can I merge this branch or will you make the MySQL virtual filter soon ?

gentledepp:
I am not sure when I can do the MySql provider - I have very little experience with MySql.
But I'll try my best.
We should always add all features to all providers, or API users will be very confused.

gentledepp:
@Mimetis 
I have to revert my decision:
I just tried to implement those missing tests for virtual syncfitlers to mysql.

Doing that I found out, that there are no MySqlSyncFiltersTests yet, where the server itself is running MySql.

So I added them thinking that everything should work,  but it doesn't.
So I added another PR #87

It does not work yet - still so many exceptions. I would appreciate your help though, as I am an absolute MySql N00B!

So from my point of view, you can merge this branch now, as the basic syncfilter feature isn't provided for MySQL anyways. :-|

Issue #90, opened 2018-08-29
[Enhancement] - allow to execute custom code upon provisioning and deprovisioning


gentledepp:
As you might already be aware of, we are creating another table (syncfilter) next to the others (base and tracking). This should happen when dotmim.sync provisions the database.

So we will add virtual methods to coreprovider, that allow to execute custom code when the scope is provisioned or deprovisioned.

You can then customize the provisioning process like so:

```

            public class CustomSqlSyncProvider : SqlSyncProvider
    {
        public CustomSqlSyncProvider(string connectionString) : base(connectionString)
        {
        }

        public CustomSqlSyncProvider(SqlConnectionStringBuilder builder) : base(builder)
        {
        }

        /// <summary>
        /// Called whenever a sync configuration has been deprovisioned
        /// </summary>
        /// <param name="args">contains information about the deprovisioning process</param>
        /// <returns></returns>
        protected override Task OnDeprovisionedAsync(DatabaseDeprovisionedEventArgs args)
        {
            return Task.CompletedTask;
        }

        /// <summary>
        /// Called whenever a table has been deprovisioned
        /// </summary>
        /// <param name="args">Provides information about the table that was deprovisioned</param>
        /// <returns></returns>
        protected override Task OnTableDeprovisionedAsync(TableDeprovisionedEventArgs args)
        {
            return Task.CompletedTask;
        }

        /// <summary>
        /// Called whenever a table has been provisioned
        /// </summary>
        /// <param name="args">Provides information about the table that was provisioned</param>
        /// <returns></returns>
        protected override Task OnProvisionedAsync(DatabaseProvisionedEventArgs args)
        {
            return Task.CompletedTask;
        }

        /// <summary>
        /// Called whenever a sync configuration has been provisioned
        /// </summary>
        /// <param name="args">contains information about the provisioning process</param>
        /// <returns></returns>
        protected override Task OnTableProvisionedAsync(TableProvisionedEventArgs args)
        {
            return Task.CompletedTask;
        }
    }

```

PR #91, opened 2018-08-29
#90 - added virtual methods that allow to customize provisioning and …

gentledepp:
…deprovisioning

fixing  #90

### DIFF
--- a/Projects/Dotmim.Sync.Core/CoreProvider.Database.cs
+++ b/Projects/Dotmim.Sync.Core/CoreProvider.Database.cs
@@ -62,6 +62,8 @@ namespace Dotmim.Sync
                             // On purpose, the flag SyncProvision.All does not include the SyncProvision.Table, too dangerous...
                             if (provision.HasFlag(SyncProvision.Table))
                                 builder.DropTable(connection, transaction);
+
+                            await OnTableDeprovisionedAsync(configuration, provision, dmTable, connection, transaction);
                         }
 
                         if (provision.HasFlag(SyncProvision.Scope) || provision.HasFlag(SyncProvision.All))
@@ -70,6 +72,9 @@ namespace Dotmim.Sync
                             if (!scopeBuilder.NeedToCreateScopeInfoTable())
                                 scopeBuilder.DropScopeInfoTable();
                         }
+
+                        await OnDeprovisionedAsync(configuration, provision, connection, transaction);
+
                         transaction.Commit();
                     }
                 }
@@ -140,7 +145,12 @@ namespace Dotmim.Sync
                             if (provision.HasFlag(SyncProvision.StoredProcedures) || provision.HasFlag(SyncProvision.All))
                                 builder.CreateStoredProcedures(connection, transaction);
 
+                            await OnTableProvisionedAsync(configuration, provision, dmTable, connection, transaction);
+
                         }
+
+                        await OnProvisionedAsync(configuration, provision, connection, transaction);
+
                         transaction.Commit();
                     }
 
@@ -158,7 +168,7 @@ namespace Dotmim.Sync
                     connection.Close();
             }
         }
-
+        
         /// <summary>
         /// Be sure all tables are ready and configured for sync
         /// the ScopeSet Configuration MUST be filled by the schema form Database
@@ -244,6 +254,60 @@ namespace Dotmim.Sync
         }
 
         /// <summary>
+        /// Called whenever a sync configuration has been deprovisioned
+        /// </summary>
+        /// <param name="configuration"></param>
+        /// <param name="provision">shows which kind of deprovisioning takes place </param>
+        /// <param name="connection">the connection used for deprovisioning</param>
+        /// <param name="transaction">the transaction used for deprovisioning</param>
+        /// <returns></returns>
+        protected virtual Task OnDeprovisionedAsync(SyncConfiguration configuration, SyncProvision provision, DbConnection connection, DbTransaction transaction)
+        {
+            return Task.CompletedTask;
+        }
+
+        /// <summary>
+        /// Called whenever a table has been deprovisioned
+        /// </summary>
+        /// <param name="configuration"></param>
+        /// <param name="provision">shows which kind of deprovisioning takes place </param>
+        /// <param name="dmTable">the affected table</param>
+        /// <param name="connection">the connection used for deprovisioning</param>
+        /// <param name="transaction">the transaction used for deprovisioning</param>
+        /// <returns></returns>
+        protected virtual Task OnTableDeprovisionedAsync(SyncConfiguration configuration, SyncProvision provision, DmTable dmTable, DbConnection connection, DbTransaction transaction)
+        {
+            return Task.CompletedTask;
+        }
+
+        /// <summary>
+        /// Called whenever a sync configuration has been provisioned
+        /// </summary>
+        /// <param name="configuration"></param>
+        /// <param name="provision">shows which kind of provisioning takes place </param>
+        /// <param name="connection">the connection used for provisioning</param>
+        /// <param name="transaction">the transaction used for provisioning</param>
+        /// <returns></returns>
+        protected virtual Task OnProvisionedAsync(SyncConfiguration configuration, SyncProvision provision, DbConnection connection, DbTransaction transaction)
+        {
+            return Task.CompletedTask;
+        }
+
+        /// <summary>
+        /// Called whenever a table has been provisioned
+        /// </summary>
+        /// <param name="configuration"></param>
+        /// <param name="provision">shows which kind of provisioning takes place </param>
+        /// <param name="dmTable">the affected table</param>
+        /// <param name="connection">the connection used for provisioning</param>
+        /// <param name="transaction">the transaction used for provisioning</param>
+        /// <returns></returns>
+        protected virtual Task OnTableProvisionedAsync(SyncConfiguration configuration, SyncProvision provision, DmTable dmTable, DbConnection connection, DbTransaction transaction)
+        {
+            return Task.CompletedTask;
+        }
+
+        /// <summary>
         /// Adding filters to an existing configuration
         /// </summary>
         private void AddFilters(ICollection<FilterClause> filters, DmTable dmTable, DbBuilder builder)
@@ -264,6 +328,5 @@ namespace Dotmim.Sync
             }
 
         }
-
     }
 }
--- a/Projects/Dotmim.Sync.Core/CoreProvider.Database.cs
+++ b/Projects/Dotmim.Sync.Core/CoreProvider.Database.cs
@@ -150,7 +150,7 @@ namespace Dotmim.Sync
 
                         }
 
-                        await OnProvisionedAsync(new DatabaseProvisionedEventArgs(provision, connection, transaction));
+                        await OnProvisionedAsync(new DatabaseProvisionedEventArgs(provision, configuration.Schema, connection, transaction));
 
                         transaction.Commit();
                     }
@@ -236,7 +236,7 @@ namespace Dotmim.Sync
                         var afterArgs = new DatabaseAppliedEventArgs(this.ProviderTypeName, context.SyncStage, script.ToString());
                         this.TryRaiseProgressEvent(afterArgs, this.DatabaseApplied);
 
-                        await OnProvisionedAsync(new DatabaseProvisionedEventArgs(SyncProvision.All, connection, transaction));
+                        await OnProvisionedAsync(new DatabaseProvisionedEventArgs(SyncProvision.All, message.Schema, connection, transaction));
 
                         transaction.Commit();
                     }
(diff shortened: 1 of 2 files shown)


gentledepp:
Well, in our case I do not want to use eventhandlers, as they do not support for awaiting asynchronous operations, which is what we require.

Aside from that, they are a frequent source of memoryleaks.

But having additional events for this case is not a bad idea I think.

Mimetis:
Make sense   

So in the opposite, can we imagine to *replace* all the events by virtual async methods ?

gentledepp:
yes, that would be an option.
But there is a downside to this:

You can add the event handlers to any implementation of coreprovider.
To override the virtual methods, however, you'd need to derive from a specific provide (e.g. SqlSyncProvider)

What would be, if we defined an interceptor for this (similar to what is done with the `ExpressionVisitor`s in .net framework)

So there would be the possibility to pass an `IEnumerable<ISyncEventInterceptor>` (or however you choose to call it) to the constructor of the CoreProvider (or even a property of it)

Whe could then provide a `SyncEventInterceptorBase` which implements the interface as virtual methods that do nothing.
So then you can go and override the methods you are interested in.

That would be "composition over inheritance" which is quite a nice concept I think.

What is your angle on this?

Mimetis:
I'm seriously thinking to change the way we handle events today...   
It's too complicated and as you said `EventHandlers` can be source of memory leaks, and theyr are not *awaitable*  
And in the opposite using a virtual method is not good since we need to create a new provider deriving from a specific provider.

I'm currently looking on the **visitor** pattern, and I want to go further on this way.  
Can you explain how you see the solution using Visitor / Interceptor ?  

That could be a quite interesting architecture discussion :)

gentledepp:
quite simple:
Provide an interface `ISyncEventInterceptor` which has all the aforementioned events defined as awaitable methods.

Now add a property `public IList<ISyncEventInterceptor> Interceptors {get;}` to the core provider.
And wherever you fire an event today, just run
```

foreach(var interceptor in Interceptors){
    await interceptor.TheAwaitableCallbackMethod(...);
}

```

This has the following pros:
* you do not have the risk of memory leaks as you would have with events
* you can make the callbacks async
* the api user can loosely configure the interceptors (composition over inheritance)
   also, he can explicitly specify their order, which may also be relevant at some point.


Enhancement:
As you may only be interested in one specific event, you do not want to implement the complete interface.
Therefore provide a SyncEventInterceptorBase class which implements these as virtual methods which just return `Task.CompletedTask`


Does that make sense to you?

Mimetis:
Here is a first proposition:

First, I use the `BaseProgressEventArgs` as base class we need to pass. This class already exists and is used in all event args. It contains some basic informations.  
Then args like `BeginSessionEventArgs`, `EndSessionEventArgs` derive from `BaseProgressEventArgs` …

So first thing, I've defined a quite simple *interface* exposing an `Intercept` method:
``` csharp
public interface ISyncInterceptor<T> where T : BaseProgressEventArgs
{
    Task Intercept(T args);
}
```
Then here is a first implementation coded in `CoreProvider`:

``` csharp
/// <summary>
/// Registering all interceptors, for each Sync stage.
/// Using a dictionary to be able to have multiple interceptors for each stage
/// We can have 0 interceptor for 'begin session stage'
/// Then having 1 inteceptor for 'end session stage'
/// Then having 4 interceptors for 'table change applied stage'
/// </summary>
private readonly Dictionary<SyncStage, List<object>> dictionary =
    new Dictionary<SyncStage, List<object>>();

/// <summary>
/// Get a list of interceptors for a dedicated sync stage
/// We can have multiple interceptors for each stage
/// </summary>
private List<object> GetInterceptors(SyncStage syncStage)
{
    if (this.dictionary.ContainsKey(syncStage))
        return this.dictionary[syncStage];

    var lstInterceptors = new List<object>();
    this.dictionary.Add(syncStage, lstInterceptors);

    return lstInterceptors;
}

/// <summary>
/// Adding an interceptor for a specific sync stage
/// I don't use a switch because an interceptor could be potentially added to several sync stage
/// </summary>
public void AddInterceptor(object interceptor)
{
    if (interceptor is ISyncInterceptor<BeginSessionEventArgs>)
    {
        var inspectors = GetInterceptors(SyncStage.BeginSession);
        inspectors.Add(interceptor);
    }

    if (interceptor is ISyncInterceptor<EndSessionEventArgs>)
    {
        var inspectors = GetInterceptors(SyncStage.EndSession);
        inspectors.Add(interceptor);
    }
}
```

Again, in `CoreProvider`, in each stage, we can call all the interceptors. For example, in the `BeginSessionAsync`:
``` csharp
public virtual Task<(SyncContext, SyncConfiguration)> BeginSessionAsync(SyncContext context, MessageBeginSession message)
{
    try
    {
        // bla bla bla code

        // Set stage
        context.SyncStage = SyncStage.BeginSession;

        // Event progress
        var progressEventArgs = new BeginSessionEventArgs(this.ProviderTypeName, context.SyncStage, null, null);

        // Not needed, now using interceptors
        // this.TryRaiseProgressEvent(progressEventArgs, this.BeginSession);

        var lstArgs = this.GetInterceptors(SyncStage.BeginSession);
        lstArgs.ForEach(bpea => ((ISyncInterceptor<BeginSessionEventArgs>)bpea).Intercept(progressEventArgs));

        return Task.FromResult((context, message.Configuration));
    }
    catch (Exception ex)
    {
        throw new SyncException(ex, SyncStage.BeginSession, this.ProviderTypeName);
    }
}
```

Now the creation of the interceptor itself:

``` csharp
/// <summary>
/// A straightforward interceptor
/// This interceptor will intercept 2 stages : Begin session and End session
/// </summary>
public class BaseSyncInterceptor : 
    ISyncInterceptor<BeginSessionEventArgs>,
    ISyncInterceptor<EndSessionEventArgs>
{
    public async Task Intercept(BeginSessionEventArgs args)
    {
        args.Action = ChangeApplicationAction.Continue;
        await Task.CompletedTask;
    }

    public async Task Intercept(EndSessionEventArgs args)
    {
        args.Action = ChangeApplicationAction.Continue;
        await Task.CompletedTask;
    }
}
```

Then add the interceptor to your `CoreProvider` using the `AddInterceptor` method.

### PROS / CONS:

**PROS**:

With this method, we don't have to fill ALL the interfaces need by `Dotmim.Sync`. In the same way, no need to create a base class who implements by default all these interfaces.   
In the same idea, I love the idea to pass 0, 1 or more intereceptors, and each interceptor could potentially handle several sync stage.

For instance, this code is equivalent to the last one:
``` csharp
    public class BeginSessionSyncInterceptor : ISyncInterceptor<BeginSessionEventArgs>
    {
        public async Task Intercept(BeginSessionEventArgs args)
        {
            args.Action = ChangeApplicationAction.Continue;
            await Task.CompletedTask;
        }
    }

    public class EndSessionSyncInterceptor : ISyncInterceptor<EndSessionEventArgs>
    {
        public async Task Intercept(EndSessionEventArgs args)
        {
            args.Action = ChangeApplicationAction.Continue;
            await Task.CompletedTask;
        }
    }
```

**CONS**:

It's quite complicated, in a sense, if we want to handle something, we NEED to create a new class somewhere.   
I mean, using the `AddInterceptor()` method requires an interface. Would be cool to simplify this part, in my opinion

In my implementation, I use several times boxing to object due to language limitations, that's not cool for performance issue, even if performances expectation are not high :)

What do you think ?
I know the visitor pattern, but I never used it. This is a first try, not sure my implementation is correct :)

Sebastien

gentledepp:
Hi!

I like your idea of not having one interface to implement all sync events, but rather use the event arg classes as generic parameters. Cudos.

Here are my remarks
1) as the `Intercept` method is async, it should normally be postfixed with "Async" => `Task InterceptAsync(...)`  - I personally do not care
2) am I correct in the assumption, that you forgot to await the interceptor calls in your `.ForEach` Extension method? 
Because the method `BeginSessionAsync` still uses `Task.FromResult(...)`!!

3) In case you do NOT want to create an interface each time, you could add an overload for a adding a func and use a wrapper class (`InterceptorWrapper<T>`)
```
    public class BaseProgressEventArgs { }
    public class BeginSessionEventArgs : BaseProgressEventArgs { }
    public class EndSessionEventArgs : BaseProgressEventArgs { }
    
    public interface ISyncInterceptor<T> where T : BaseProgressEventArgs
    {
        Task Intercept(T args);
    }

    public class X
    {
        private class InterceptorWrapper<T> : ISyncInterceptor<T> where T : BaseProgressEventArgs
        {
            private readonly Func<T, Task> _interceptor;

            public InterceptorWrapper(Func<T,Task> interceptor)
            {
                _interceptor = interceptor;
            }

            public Task Intercept(T args)
            {
                return _interceptor(args);
            }
        }

        public void AddInterceptor<T>(Func<T, Task> interceptorFunc) where T : BaseProgressEventArgs
        {
            var interceptor = new InterceptorWrapper<T>(interceptorFunc);
            
            AddInterceptor(interceptor);
        }

        public void AddInterceptor(object interceptor)
        {
            if (interceptor is ISyncInterceptor<BeginSessionEventArgs>)
            {
                var inspectors = GetInterceptors(SyncStage.BeginSession);
                inspectors.Add(interceptor);
            }

            if (interceptor is ISyncInterceptor<EndSessionEventArgs>)
            {
                var inspectors = GetInterceptors(SyncStage.EndSession);
                inspectors.Add(interceptor);
            }
        }
    }

```

Mimetis:
* 1 : You're right :)
* 2 : You're right as well !
* 3 : Your idea is really good.

Th final call to `AddInterceptor` could be something like that:

``` csharp
this.AddInterceptor(async (BeginSessionEventArgs args) =>
{
    args.Action = ChangeApplicationAction.Continue;
    await Task.CompletedTask;
});
```

And this is quite more simple !

I'm pretty excting to be able to have multiple solutions: 
* 1 big class implementing all the interfaces we want to interact with
* Multiple classes implementing 1 interface
* A proxy class allowing us to use a `Func`

If we go further, what do you think of having more wrappers ? Let's imagine instead of having this:

```csharp
// creating a new SqlSyncProvider
// ....

// Add interceptor to handle failed changes
_sqlSyncProvider.AddInterceptor(async (BeginSessionEventArgs args) =>
{
    args.Action = ChangeApplicationAction.Continue;
    await Task.CompletedTask;
});
```

we have something like that:
``` csharp

this.InterceptBeginSession(async options =>
{
    options.Action = ChangeApplicationAction.Continue;
    await Task.CompletedTask;
});

```
Just because we wrap all events in `CoreProvider`:
``` csharp

public void InterceptBeginSession(Func<BeginSessionEventArgs, Task> options)
{
    this.AddInterceptor(options);
}

public void InterceptEndSession(Func<EndSessionEventArgs, Task> options)
{
    this.AddInterceptor(options);
}

// and so on ...

```

It's really something we see today in modern language, what we use to call **Fluent** pattern. Not sure it's exactly the same, but no too far, I guess.

My only problematic today, is this weird boxing of Object... It could awful for users if they don't understand the meaning of this...

gentledepp:
I think these more concrete methods like `InterceptBeginSession` and `InterceptEndSession` are quite a good idea, but I'd rather implement them as extension methods.

```
public static class CoreProviderExtensions {

public static void InterceptBeginSession(this CoreProvider p, Func<BeginSessionEventArgs, Task> options)
{
    p.AddInterceptor(options);
}

public static void InterceptEndSession(this CoreProvider p, Func<EndSessionEventArgs, Task> options)
{
    p.AddInterceptor(options);
}

}

```

This pattern of "define the interface as general as possible and add extension methods for the ease of use" is a best practice in my experience.

The rationale behind this is simple: Mocking gets much easier.

If you have a class/interface that has many methods, you may have to mock all them in unit tests.
If you, however, only have one generalized method (which is used by extension methods to make the api more usable to developers), you only have to mock that one method.

That is why my loggers always look like this:


```

public interface ILogger{
      void Log(Severity severityl, string tag, string message, Exception x);
}

public static class LoggerExtensions{

 public static void Info(this ILogger l, string message){
      l.Log(Severity.Info, string.Empty, message, null);
 }

 ....
 
 public static void Error(this ILogger l, string message, Exception x){
      l.Log(Severity.Error, string.Empty, message, x);
 }

 public static void Error(this ILogger l, Exception x){
      l.Log(Severity.Error, string.Empty, string.Empty, x);
 }

}

```

That being said, the `AddInterceptor<T>(Func<T,Task> interceptor)` should - in that sense - also be defined as an extension method ;-)

Does that make sense to you?

Mimetis:
Make sense!!

Well... Who will create the PR? 😂

gentledepp:
I'd be happy to do so, but have no time at the moment. I just - at least - try to write back to you in any comments.

Are all unit tests up and runnning again on the master branch?

Mimetis:
I will probably have some times next week, so I think I will handle it.

Yes, the unit tests are good on the master branch. I finally found and fix [this weird bug](https://github.com/Mimetis/Dotmim.Sync/pull/124/commits/7e59aaeb370cab2e7b8a45a5396718535474311e) I have for weeks... 

Be careful the master branch has breaking changes from the current **0.2.0** release packages

Issue #51, opened 2018-04-27
Prefix for tracking tables and stored procedures. 


marjanoh2:
Is there an option to put a prefix on the tracking tables and stored procedures ?

Mimetis:
It's an option I have considered before, but honestly, I didn't see the use cases where this option were important, so far.
Why do you need to personalize the prefix for tracking items ? Just for my knowledge :)

marjanoh2:
For better visibility. I think the prefix is a good way to separate tracking item.

Mimetis:
Ok, I see. So you need an option for both. Prefix and suffix ?

gentledepp:
👍  for this feature.
Scenario:
We have a system that uses the old microsoft sync framework.
Now we want to slowly migrate to dotmim.sync.
For that to work smoothly, we need to run both frameworks side-by-side for some time.

This is not possible, as there are name clashes (e.g. on the tracking tables)!

Mimetis:
Hey @marjanoh2 @gentledepp 
I have added a new featur for this particular scenario
You can try something like that to configure the prefix and suffix for each tables:

``` cs
        var tables = new string[] {"ProductCategory",
                "ProductDescription", "ProductModel",
                "Product", "ProductModelProductDescription",
                "Address", "Customer", "CustomerAddress",
                "SalesOrderHeader", "SalesOrderDetail" };

        var configuration = new SyncConfiguration(tables)
        {
            ScopeName = "AdventureWorks",
            ScopeInfoTableName = "tscopeinfo",
            SerializationFormat = SerializationFormat.Binary,
            DownloadBatchSizeInKB = 400,
            StoredProceduresPrefix = "s",
            StoredProceduresSuffix = "",
            TrackingTablesPrefix = "t",
            TrackingTablesSuffix = "",
        };
```

gentledepp:
that is AWESOME! Thanks! Will try that out 👍

Mimetis:
Hey @gentledepp  did you try it already ?

gentledepp:
I did not find the time yet, but will most probably be able to verify this next week!

marjanoh2:
I tested the new version and works well with default dbo tables. But it makes a problem with tables that have a different schema. For example:
Table: deimos.r_bills
Prefix for tables and procedures tr_
Gives this error:
Dotmim.Sync.SyncException: 'The specified schema name "tr_deimos" either does not exist or you do not have permission to use it.'

[...]

PR #93, opened 2018-08-30
Fix for #51 - when using webproxy, pre/postfixes of triggers are ignored

gentledepp:
Setting up trigger pre- and postfixes using a webproxy (asp.net) does not work.
The pre- and postfixes are ignored on the server side (verifyable with SqlServer provider)

Completes issue #51


Mimetis:
omg, I did not check the surrogate :(
Thx for the correction

PR #96, opened 2018-09-07
Automagically creating nuget packages along with symbol- and source- …

gentledepp:
…package on each release build

as requested in #76

How do you like that?


Mimetis:
Seems to be a good idea ;)

Mimetis:
will be available on the next release !

PR #99, opened 2018-09-11
Mimetis filter clause

Mimetis:
Hi guyz

Need your help on the **filter** behavior. 

_But first of all, take a chair, a coffee, turn on spotify on your best music of the day, and breath..._ 

Today we are able to filter on table that have a direct link to the parent table.   
But if we habe a 2nd (or more) deep level of filter it's impossible to filter values (or you have to duplicate the key on all tables involved in the filter)

_This issue is a discussion, I'm writing what I actually think and a lot of things could changed over time :)_

## Example (of failure)

Let's imagine this scenario : We have a `Customer` table who have `Address` through a `CustomerAddress` many to many relations.

Those customers are linked to `Employee`  (each customer has one employee and one employee could potentially have several customers)

> **Employee** :                       **EmployeeID** / FirstName / LastName
> |-> **Customer** :                 **CustomerID** / **EmployeeID** / FirstName / LastName
> |---->**CustomerAddress** : **CustomerID** / **AddressID**
> |------->**Address** :            **AddressID** / City / State / PostalCode
> 

The SQL Tables (simplified) could be something like that:

``` sql
CREATE TABLE [dbo].[Employee](
	[EmployeeId] [int] IDENTITY(1,1) NOT NULL,
	[FirstName] [nvarchar](50) NOT NULL,
	[LastName] [nvarchar](50) NOT NULL)

CREATE TABLE [dbo].[Customer](
	[CustomerID] [uniqueidentifier] NOT NULL,
	[EmployeeID] [int] NOT NULL,
	[FirstName] [nvarchar](50) NOT NULL,
	[LastName] [nvarchar](50) NOT NULL)

CREATE TABLE [dbo].[Address](
	[AddressID] [int] IDENTITY(1,1) NOT NULL,
	[AddressLine1] [nvarchar](max) NOT NULL,
	[City] [nvarchar](30) NOT NULL)

CREATE TABLE [dbo].[CustomerAddress](
	[CustomerID] [uniqueidentifier] NOT NULL,
	[AddressID] [int] NOT NULL,
	[AddressType] [nvarchar](50) NOT NULL)

```

## How it works (and fails) today

**Imagine we want to synchronization, per device, only customers belonging to one employee** .

Today, we can do something like that, in `Dotmim.Sync`:

```csharp
// On the server side, we had the filter
var filterClause1 = new FilterClause("Employee", "EmployeeID");
var filterClause2 = new FilterClause("Customer", "EmployeeID");

// On the client side, we had the SyncParameter for the device 
// Let's say we want only employee with ID = 1
var paramEmployee = new SyncParameter("Employee", "EmployeeID", 1)
var paramCustomer = new SyncParameter("Customer", "EmployeeID", 1)

```
It's fine for the `Employee` table and also for the `Customer` table.

But it won't work for the `Address` and `CustomerAddress`.
Eventually it will fail, because we should have in the `CustomerAddress` some rows that are referencing employees that are not present because of the filter.

## A solution to discuss

Ok, so we want to be able to sync all the tables linked (more or less) to one employee.

We have to refactor (and introduce some breaking changes, I guess) the `FilterClause` and `SyncParameter` classes, and the **SQL** generation as well.

### SyncParameter
In my mind, we should be able to inject a parameter, without any verification on its corresponding column on any table. This will allow us to create some **Virtual Parameter**

Since we don't have any resolution based on the name, this parameter should be:
- Typed on server side
- Just a Name / Value on client

On server side, let's say we can use a `DmColumn` to create this parameter, and use all the `DmColumn` properties to precise length, unicode, null, precision, scale ....
This column will be part of all `FilterClause` which need it

``` csharp
// Create the parameter that will be in the stored procedure parameter
var pEmpId = DmColumn.CreateColumn("empId", typeof(Int32));
pEmpId.AllowDbNull = true;
```
On client side, just set a `SyncParameter` with the correct name and its value

``` csharp
// We want all rows for the employee with EmployeeId = 1
var syncParameter new SyncParameter("empId", 1)
```

This parameter will generate a **SQL** variable, like `@empId int = null`

### FilterClause

Today the `FilterClause` creates a parameter that has the same name as the column name we want to filter.   
```csharp
var filterClause1 = new FilterClause("Employee", "EmployeeID");
```
Now that we have a `DmColumn`, with no restriction, the new `filterClause` will be something like

``` csharp
// create the parameter
var pEmpId = DmColumn.CreateColumn("empId", typeof(Int32));
// create the filter clause
var customerFilter = new FilterClause(pEmpId);
```
this `customerFilter` will generate something like that, in the stored procedure declaration:

```sql
ALTER PROCEDURE [dbo].[Customer_empId_selectchanges]
	@sync_min_timestamp bigint,
	@sync_scope_id uniqueidentifier,
	@sync_scope_is_new bit,
	@sync_scope_is_reinit bit,
	@empId int = null
```
Now we have the in-parameter, we need to say to the `FilterClause` which table we want to filter.

I tried to get the most intuitive solution, and I guess, using some sort of **Fluent API** should work.

Let's say we want to **FILTER** the `Customer` table **ON** the `EmployeeId` column. Using a **Fluent API** style, we could do something like
```csharp
customerFilter.Filter("Customer").On("EmployeeId");
```

Going forward, for a 2nd level table, let's say for `CustomerAddress`, we want to **FILTER** `CustomerAddress` which **JOIN** `Customer` that will filter **ON** then `EmployeeId` column:
```csharp
custAddFilter.Filter("CustomerAddress")
                     .Join("Customer")
                     .On("EmployeeId");

```
And then, eventually for the last level `Address`, we can do something like:

``` csharp
addressFilter1.Filter("Address")
                       .Join("CustomerAddress")
                       .Join("Customer")
                       .On("EmployeeId");
```

## Final sample code

Here is my final code, I want to be able to implement, which will filter all **Customers** and all **Orders** related to one filtered **Employee**
```csharp
// Create the parameter that will be in the stored procedure parameter
var pEmpId = DmColumn.CreateColumn("empId", typeof(Int32));
pEmpId.AllowDbNull = true;

// Filter table Customer on EmployeeID = @empID
FilterClause customerFilter = new FilterClause(pEmpId);
customerFilter.Filter("Customer")
                .On("EmployeeId");

// Filter on CustomerAddress that will Join Customer that will filter on EmployeeID = @empID
FilterClause customerAddressFilter = new FilterClause(pEmpId);
customerAddressFilter.Filter("CustomerAddress")
                    .Join("Customer")
                    .On("EmployeeId");

// 1st Filter on Address that will Join CustomerAddress that will join Customer that will filter on EmployeeID = @empID
FilterClause addressFilterOnCustomer = new FilterClause(pEmpId);
addressFilterOnCustomer.Filter("Address")
                .Join("CustomerAddress")
                .Join("Customer")
                .On("EmployeeId");

// 2nd Filter on Addres that will join EmployeeAddress that will filter on EmployeeID = @empID
FilterClause addressFilterOnEmployee = new FilterClause(pEmpId);
addressFilterOnEmployee.Filter("Address")
                .Join("EmployeeAddress")
                .On("EmployeeId");


// Filter on Employee that will filter on EmployeeID = @empID
FilterClause2 employeeFilter = new FilterClause2(pEmpId);
employeeFilter.Filter("Employee").On("EmployeeId");

// Filter on EmployeeAddress that will filter on EmployeeID = @empID
FilterClause2 employeeAddressFilter = new FilterClause2(pEmpId);
employeeAddressFilter.Filter("EmployeeAddress").On("EmployeeId");

// Filter on SalesOrderHeader that will Join Customer that will filter on EmployeeID = @empID
FilterClause2 sohFilter = new FilterClause2(pEmpId);
sohFilter.Filter("SalesLT.SalesOrderHeader")
                .Join("Customer")
                .On("EmployeeId");

// Filter on SalesOrderDetails that will Join SalesOrderHeader Customer that will filter on EmployeeID = @empID
FilterClause2 sodFilter = new FilterClause2(pEmpId);
sodFilter.Filter("SalesLT.SalesOrderDetail")
                .Join("SalesLT.SalesOrderHeader")
                .Join("Customer")
                .On("EmployeeId");

// Customer , CustomerAddress, Address
Agent.Configuration.Filters.Add(customerFilter);
Agent.Configuration.Filters.Add(customerAddressFilter);
Agent.Configuration.Filters.Add(addressFilterOnCustomer);

// Employee, EmployeeAddress, Address
Agent.Configuration.Filters.Add(employeeFilter);
Agent.Configuration.Filters.Add(employeeAddressFilter);
Agent.Configuration.Filters.Add(addressFilterOnEmployee);

// SalesOrderHeader, SalesOrderDetail
Agent.Configuration.Filters.Add(sohFilter);
Agent.Configuration.Filters.Add(sodFilter);


```

#### Breaking changes
This change coud introduce breaking changes in your code

#### Foreign keys constraints
Since we don't need to precise the relationships in the **Fluent API** all joins will be based on the foreign keys constraints from the server database.   
So It can't work without them.

#### Tracking tables
To be able to retrieve deleted rows through all the tables level , I must use the tracking tables.   
So all tracking tables will have new columns : all foreign key columns.   
For example, the `Customer` table has a foreign key to `Employee`, so the tracking table will look like:
``` sql
CREATE TABLE [dbo].[Customer_tracking](
	[CustomerID] [uniqueidentifier] NOT NULL,
	[create_scope_id] [uniqueidentifier] NULL,
	[update_scope_id] [uniqueidentifier] NULL,
	[create_timestamp] [bigint] NULL,
	[update_timestamp] [bigint] NULL,
	[timestamp] [timestamp] NULL,
	[sync_row_is_tombstone] [bit] NOT NULL,
	[last_change_datetime] [datetime] NULL,
	[EmployeeID] [int] NULL,
```

#### TSQL generation

TSQL generation will be complex for all `_SelectChanges` stored procedures, involving a lot of relations, and maybe a `UNION` when we had multiple `FilterClause` on the same table

#### Do we have to create the filters on the **clients** databases ? 
Today, the filters are created on both side, essentially because the code on server and on client is exactly the same.   
Introducing this kind of exceptions could have a huge impact on the overall architecture

#### Sql generation on client
In the other side, if we allow injecting **SQL** in the stored procedures, well we should make a SQL generation for each of the clients (the SQL for SQL Server is not the same as the SQL used in SQLite for instance)

#### What about Virtual Filters
I did not (yet) study the new virtual filters introduce by @gentledepp.  
Do you think it's compatible ?

Guyz, any feedback will be a gift :)


gentledepp:
Hi! Sorry for the late reply - I am working long hours the past few weeks 

1) Virtual filters are just filters which do not map to any table column directly, but are just there so that dotmim.sync passes them to the `select_changes` stored procedure. This is needed for us so we can modify the procedure to use some more advanced filtering techniques.
2) What will also be required is constant filters. These are the opposite of virtual filters: They just filter a table by a constant filter expression, so they won't get any parameters passed in. This is needed for legacy scenarios. For example: We have some table "Pictures" which have some column "Type". We only want to sync type "A" of the pictures (a constant `where Type = 'A'` SQL expression). However, on the client side, we do not want this column anymore (it is obsolete, but needs to remain there to support the old sync scenario with *Microsoft Sync Framework*) hence we cannot pass any filter.

As for your proposed *table join filter* scenario, I must say that we have had quite bad experiences with this approach.
First of all, it really only works on one-to-many scenarios, not in many-to-many. (At least we did not get it to work)

Let's assume you want to sync all data for an employee.
In that case, you'd pass the `@employeeId` parameter to the `SyncAgent` and in the `Customer_EmployeeId_selectchanges` stored procedure on the server side, you will just use this filter.
Now this works, *as long as the EmployeeId column in table Customer does not change*. 
So when some user updates the `[Customer]` record to belong to another employee, your filter won't catch this change.
In order to do that, you'd not only need to join to the '[Customer]' table, but also with '[Customer_tracking]'.

Now that even gets worse if you want to sync a whole table hierarchy, as described in your example:`
`[Employee]`->`[Customer]`->`[CustomerAddress]`->`[Address]`

Mimetis:
> 1) Virtual filters are just filters which do not map to any table column directly, but are just there so that dotmim.sync passes them to the `select_changes` stored procedure. 

That's what is done, in my example, here:

```csharp
// virtual filter
var region = new DmColumn<String>("region")
{
    AllowDBNull = true,
    DbType = DbType.StringFixedLength,
    MaxLength = 12
};
FilterClause2 regionSqlFilter = new FilterClause2(region);
regionSqlFilter.Filter("SalesLT.Product");
```
Because you don't add the `On(Colmun_To_Match)` after the `Filter(Table)` method, the `region` parameter will be generated in the `SalesProduct` `SelectChanges` stored procedure, but won't be used in it.

> 2) What will also be required is constant filters. These are the opposite of virtual filters: They just filter a table by a constant filter expression, so they won't get any parameters passed in. This is needed for legacy scenarios. For example: We have some table "Pictures" which have some column "Type". We only want to sync type "A" of the pictures (a constant `where Type = 'A'` SQL expression). However, on the client side, we do not want this column anymore (it is obsolete, but needs to remain there to support the old sync scenario with *Microsoft Sync Framework*) hence we cannot pass any filter.

In this new Filter framework, you will be able to do this kind of things.
- Add a virtual parameter with a default value 
- Add the custom `Where` clause

```csharp
var regionType = new DmColumn<Int>("regionType")
{
    AllowDBNull = true,
    DbType = DbType.Int,
    DefaultValue = 4
};
FilterClause2 regionSqlFilterType = new FilterClause2(regionType);
regionSqlFilterType.Filter("SalesLT.Product").OnCustom("[SalesProduct].[RegionType] = @regionType
```

> As for your proposed *table join filter* scenario, I must say that we have had quite bad experiences with this approach.
> First of all, it really only works on one-to-many scenarios, not in many-to-many. (At least we did not get it to work)

Yes I had a hard time with many-to-many filters :)  
Eventually it works, thanks to `Union` queries.  In my sample the relation beetween `Customer` and `Address` is a many-to-many relationship with `CustomerAddress` in beetween.

 
> Let's assume you want to sync all data for an employee.
> In that case, you'd pass the `@employeeId` parameter to the `SyncAgent` and in the `Customer_EmployeeId_selectchanges` stored procedure on the server side, you will just use this filter.
> Now this works, *as long as the EmployeeId column in table Customer does not change*. 
> So when some user updates the `[Customer]` record to belong to another employee, your filter won't catch this change.
> In order to do that, you'd not only need to join to the '[Customer]' table, but also with '[Customer_tracking]'.

This is what is done, here. All joined relationships, in the filters clauses, are made based on the `_Tracking` table.


> Now that even gets worse if you want to sync a whole table hierarchy, as described in your example:
> `[Employee]`->`[Customer]`->`[CustomerAddress]`->`[Address]`
> And worst of all, it is **terribly slow...**

I imagine that we can make the indexes better.

> That is, why we took a totally different apporach with https://github.com/gentledepp/SyncFilters
> There, we keep a row for each record that should be syned (per sync table)
> So for each table, you not only have a `_tracking` table, but also a `_syncfilters` table which basically contains a row for each user that shall recieve this record.
> Now we did some optimizations so we do not have to really store a record per user, but can group users and just add one record for that group, but you can read about all that in the sample repository that I linked above.
 
This may be a good solution. 

I have look your solution, especially your `SQL` queries.  
It's quite complicated, and in the `SelectChanges` you are using relations to existing tables (`principaledge `), so I guess, it's not generic

SO....

We have a problem in filter column changed value 
The problem is still the same, even with this new solution. 
If we change the filter column (`EmployeeId`) it will not be deleted in the client database, because the filter will not be applied.
I guess, we can do something with a new tracking table, but this is not a solution which inspires me a lot... I really want to keep the additionnal tables, as low as possible, but I will make a test.

gentledepp:
Ok, I see! :-)

Our *syncfilter table approach* is not intended to be part of Dotmim.Sync, but I could imagine building it as an extension to the framework in the future.

The `PrincipalEdge` table is a way to store users, groups and their relationships as a graph.
This is taken from the following code project articles:
[A Model to Represent Directed Acyclic Graphs (DAG) on SQL Databases](https://www.codeproject.com/Articles/22824/A-Model-to-Represent-Directed-Acyclic-Graphs-DAG-o)
[A Fairly Capable Authorization Sub-System with Row-Level Security Capabilities (AFCAS)](https://www.codeproject.com/Articles/30380/A-Fairly-Capable-Authorization-Sub-System-with-Row)
As this approach stores the whole *transitive closure* of the graph into the table, it is uses quite a lot of space, but using it requires just one simple join (as opposed to a recursive CTE) so it is really, really fast.

One drawback of our approach is, that you need to mainain the syncfilter table in your application code. But as we do have a CQRS-like infrastructure (with a write model and an asynchronous process which updates the read model) it is not too hard for us to maintain.

One huge benefit is, that we can use it to define which user gets access to data based on his/her permission level.
To do that, we just add another user to a group (add a row to the principaledge table). That causes all the data, that this user should be able to see now, to be synced to his/her device. And more importantly, it does not affect any other users at all.

Also, we can apply a time-based filter scenario, that would not be possible with dotmim.sync filters only: For example, we specify that cancelled or completed tasks remain on a users device for 14 days only. This is simply done by recalculating the sync-rules every once in a while using some background process.

So I think our solution is not a *replacement* for what you plan to build, but an *addition*.
Does that make sense to you?

Mimetis:
Yes, makes senses.

I know we are not in a V1 product, and breaking changes could happen until we reach a stable version, but I just wanted to know if I can safely merge this PR, and start a new build, based on it

gentledepp:
Regarding the virtual filter approach, that we contributed, I have one humble request:
We are under a lot of pressure right now and will probably release a beta in 2 weeks. 
Could you please make sure that your changes are downwardly compatible with the *virtual filters*? I do not want to maintain a separate repository of Dotmim.Sync for the sake of our product if it is not absolutely necessary.
If there are any changes required, we'd need to do that in the next 2-3 weeks or so. Is that acceptable for you? :-|

Mimetis:
Maybe you can try this PR, locally, and gives your feedback ?

gentledepp:
sure I will do, but I won't be able to do so until next week.
So for the next 2 weeks, we can easily change **everything** - don't you worry about that.

Mimetis:
I can wait until next week :)

gentledepp:
Hi! Finally, I had some time to think this through a bit...

my feedback:

### Breaking Changes
Adding the DmColumn capabilities to the FilterClause is a good idea, for backward compatibility, we should keep the old `new FilterClause(string, string)`

### Tracking tables
I hope, that if we do not use joins, the foreign key columns will *not* be part of the tracking table, right? Because that would mean unnecessary storage waste

### Foreign key constraints
As I understand, you can *never* join without foreign key constraints?
```

	custAddFilter.Filter("CustomerAddress")
                     .Join("Customer")
                     .On("EmployeeId");


```	
Is it really necessary to provide the column name "EmployeeId"  here? As you use the foreign key constraints to navigate to the "Customer" table, you can *infer* the key column, right?

Note: Personally, I do not care if only foreign key constraint - joins are supported. I guess it is easier to maintain on your side.
					   

### Do we have to create the filters on the client's databases?
No, you don't - because that would ONLY be required in a peer to peer scenario where all systems are equal.
but as Dotmim.Sync only supports the client-server type where the server is always the single source of truth, the only thing you'll ever want to use filters for is to reduce the amount of data exposed to the clients (for security and scalability reasons)

### What about Virtual Filters
your *virtual parameter* is exactly the same as a *virtual filter*



### Missing in my opinion
 -[ ] the ability to filter for a constant expression
	e.g. in a legacy system scenario, there are multiple different user records in the table, you only want to sync the ones where [IsEmployee] = 1, as you create some employee management app (bad example - sry)
	* this filter is constant=> no parameter should be required on the client side.
	* the column value of [IsEmployee] will *never* change (at least not in the better/real-world scenarios that I have experienced)

Last concern:
As synchronizing data across multiple tables is very, very, very complicated, I hope
*) you tested this concept thoroughly including all possible cases of conflicts and errors (like that whenever a record leaves the scope, Dotmim.Sync still keeps track of it - as opposed to the old Microsoft Sync Framework)
*) you did some performance checks (possibly on Azure SQL as that one is very bitchy when it comes to 'costly' queries...

If you did that and are confident, then please: Go for it :-)

YannPenven:
Hello, Is this feature still worked on ? Can i do anything to help ?

Mimetis:
hello @YannPenven 
This PR is quite complex (more than I was expecting, to be honest)

Actually, this PR is "in progress", but is really outdated, compared to the master branch.   
Before moving forward, I have to merge this code from the master branch.

Once it's done, I will more than happy to get your help !

Sebastien

PR #108, opened 2018-09-20
Fix - SyncConfiguration is overriden before BatchDirectory is persisted

lordofffarm:
Also, do not create BatchDirectory on client, if BatchMode is off.


Mimetis:
Thx @lordofffarm !

Issue #118, opened 2018-11-20
[Bug] Cannot sync updated FK that refers to new row on client db


gb0o:
Steps to reproduce
- Deploy AdventureWorks db to SQL Server
- Sync to Sqlite db on client
- Create a new Product row in AdventureWorks SQL Server db
- Sync new Product to client db
- Create a new ProductCategory on client db
- Set the ProductCategoryID FK on the new Product to be the new ProductCategory on client db
- Sync back to the server
- Inspect the new Product row and observe ProductCategoryID FK is NULL on the server db. It should have been updated.

I traced the problem to the order of Update and Insert. Each of these actions occurs in table order as expected. However a FK constraint exception occurred on the server side when a FK update was executed on the Product row before the insert of the new ProductCategory row. The solution would be to execute both Update and Insert for a given table in order and not all Updates in order then all Inserts in order. 

The problem was difficult to find because everything looked correct from the client side. One row was created and one updated. However the FK constraint exception was silently eaten on the server side. A conflict was logged on the server side but this is not actually a conflict. The server side should throw the exception so that the client is notified of the problem.

@Mimetis happy to put up a PR. I have made some changes and created a test on my fork which I think will work.

Regards,
Guy

gb0o:
@Mimetis am I allowed permissions to create a branch?

Mimetis:
Hello
Sorry for late response, I'm very busy these days...
Of course, you can create a branch and push a PR !

Thx

Sebastien

gb0o:
@Mimetis here is my PR to fix this issue: https://github.com/Mimetis/Dotmim.Sync/pull/121

gb0o:
@Mimetis confirmed this issue has been fixed. When running locally temporary files at C:\Users\<user>\AppData\Local\Temp\DotmimSync are cleared after the server and client are finished exchanging data.

PR #121, opened 2018-12-17
fixed constraint when table with FK not yet inserted

gb0o:


Projects/Dotmim.Sync.Core/CoreProvider.ApplyChanges.cs
```diff
@@ -58,33 +50,46 @@ public virtual async Task<(SyncContext, ChangesApplied)> ApplyChangesAsync(SyncC
                     // -----------------------------------------------------
                     if (!message.FromScope.IsNewScope)
                     {
-                        changeApplicationAction = this.ApplyChangesInternal(context, message, connection, applyTransaction, DmRowState.Deleted, changesApplied);
-
+                        // for delete we must go from Up to Down
```
Mimetis: Why the loop through all tables is now placed here ?
I mean, before we had this loop in the `ApplyChangesInternal` method.
This code works correctly and no reason for rollback it, but I want to know why you have decide to make the loop here and not in the `ApplyChangesInternal` anymore ?

Mimetis: Ok, I answer my own question, it's because of bug #118 ...
I'm currently reviewing this PR, I try to understand why we have the issue (that we should not have even in the last version )
Thx for the PR !

Mimetis: Ok I've understand.
It's awful to see that I did not managed this scenario so far …

So thanks a lot for your PR @gb0o !! Everything seems correct, we are good to merge !

Mimetis:
Thx @gb0o for your PR, it's merged !

Projects/Dotmim.Sync.Core/CoreProvider.ApplyChanges.cs
gb0o: @Mimetis sorry for the late reply. I see you figured why the changes was needed. In my case I got a constraint violation on a row not yet inserted because all table updates were applied before all tables inserts. When updates and inserts are handled together for each table and in the correct order every thing works. The code ended up more verbose with multiple outer loops. I could possibly improve it by getting ride of the outer loops and passing a list of operations to `ApplyChangesInternal ` such as new[] {SqlCmd.Update, SqlCmd.Insert} where SqlCmd is an new enum. There might already be such a enum.

PR #135, opened 2019-01-30
Issue67

lordofffarm:
Hi, @Mimetis!
This should solve issue 67.
Tried to leave it as compatible and clean as I could.
Changed tests a little bit, but there are no tests for the concurrent synchronization.
Also, did not add anything for the 4.6.2 build.
Please, leave some feedback!

### DIFF
--- a/Projects/Dotmim.Sync.Web.Server/DependencyInjection.cs
+++ b/Projects/Dotmim.Sync.Web.Server/DependencyInjection.cs
@@ -1,50 +1,97 @@
 ﻿using Dotmim.Sync;
 using Dotmim.Sync.Web.Server;
 using System;
+using System.Runtime.CompilerServices;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Options;
+[assembly:InternalsVisibleTo("Dotmim.Sync.Tests")]
 
 namespace Microsoft.Extensions.DependencyInjection
 {
     public static class DependencyInjection
     {
+        public static bool RegisterAsSingleton;
+        
+        private static Type _providerType;
+        private static string _connectionString;
+        private static SyncConfiguration _syncConfiguration;
+
         /// <summary>
         /// Add the server provider (inherited from CoreProvider) and register in the DI a WebProxyServerProvider.
         /// Use the WebProxyServerProvider in your controller, by inject it.
         /// </summary>
-        /// <typeparam name="TProvider">Provider inherited from CoreProvider (SqlSyncProvider, MySqlSyncProvider, OracleSyncProvider) Shoud have [CanBeServerProvider=true] </typeparam>
+        /// <typeparam name="TProvider">Provider inherited from CoreProvider (SqlSyncProvider, MySqlSyncProvider, OracleSyncProvider) Should have [CanBeServerProvider=true] </typeparam>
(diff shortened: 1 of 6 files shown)


Mimetis:
Nice to use a `MemoryCache` object to handle proxy server.

Looking at your code, each client (identified by its `SessionId`) has its own proxy server running.

So do we still need the `Singleton` version of it ?  

Just a question, before making the code review, right now :)

lordofffarm:
> So do we still need the Singleton version of it ?

I don't think so, left it there for compatibility

lordofffarm:
> So I'm thinking about deleting the specific .Net FX part of web proxy server

Yes, I was thinking the same thing, that is why I didn't implement anything for that version.

Mimetis:
First thing I noticed is now we have a static method to handle the request

``` csharp
await WebProxyServerProvider.HandleRequestAsync(context, provider =>
{
}
```

So we don't use anymore any DI process (from the developer perspective) in the Controller ?   
These lines are not mandatory anymore ?

``` csharp
private WebProxyServerProvider webProxyServer;

// Injected thanks to Dependency Injection
public SyncController(WebProxyServerProvider proxy)
{
    webProxyServer = proxy;
}
```
Maybe we should merge your `SyncMemoryCache` singleton and a light `WebProxyServerProvider` just to handle the configuration object.

In that way, we should continue to be able to inject this singleton in the controller, then the `HandleRequestAsync` would be available again as a non static method

Then the object you stored, for each `SessionId`, in the  `SyncMemoryCache` (which is today a `WebProxyServerProvider` instance) could be a new one, let's call it `SessionCoreProvider`, which will handle the sync process for this particular `SessionId`

Your thought ?
I'm currently working on it today, so if you're agree, I can make a commit on your branch, and we could continue to work on this PR ?

Sebastien

Mimetis:
just commit a first version of my thoughts on it.
As you said, we don't need everything, I'm pretty sure, we can simplify a lot.

For instance, I rely only on `MemoryCache`, so need anymore of `Session`

I've create the `SyncMemoryProvider` which will handle the sync process. nothing really completed on this part, just take everything from `WebServerProxy` that is related to provider

`WebProxyServerProvider` is what we are injecting in our DI.

Thanks to `MemoryCache` (I removed `SyncMemoryCache`, since it's not very valuable) the singletong instance of `WebProxyServerProvider` will retrieve the `SyncMemoryProvider` instance related to the session id.

For now, my only problem is that this scenario works only if we are using DI in an ASP.NET application

Mimetis:
it seems I ve created an other PR ?   
Im so dumb with Git sometimes....

Mimetis:
> Ok, I just realized that we might not even need to keep track of instances of WebProxyServerProvider and any caching.
> What we could do is register LocalProvider and SyncConfiguration as Scoped and pass them as constructor arguments to WebProxyServerProvider.
> Or we could cache LocalProvider and SyncConfiguration in SyncMemoryCache singleton object

Don't forget we need to be sure a client is not requesting a new Sync until the last one is over.

That's why, for now, we cache the while provider object

lordofffarm:
Thank you @Mimetis :)

About SyncMemoryCache
> I removed SyncMemoryCache, since it's not very valuable

I did it that way, so that library users do not have to call AddMemoryCache() in Startup,
and to isolate our own singleton object, because users might want to use their own memory cache by adding it through AddMemoryCache() and use it in their own purposes.

What do you think?

PR #141, opened 2019-02-06
Added action information for table creation exceptions

Thaval:
**Care**: This is the first PR I ever made. So please double check if it's fine.
 
I only added some keywords to the exceptions thrown when creating a table, so if an exception is thrown, people can comprehend which step went wrong.

Hint: I used the wrong credentials when I committed. So this PR is not tagged to this account.


Mimetis:
thx, seems correct :)

Issue #158, opened 2019-07-21
SQLite Error 19: 'FOREIGN KEY constraint failed'.. Type:Data.


pulsegroupb2b:
We are trying to get this great sync package up and running between our sqlite clients and an azure sql server. We are getting the above error when we try and Sync.
`BeginSession:   
ScopeLoading:   Id:f733ba42-3aca-458b-b05c-e4b30680b323 LastSync: LastSyncDuration:0 SyncState:Successful
SchemaApplied:  Tables count:8
DatabaseTableApplied:   TableName: ScanData Provision:All
DatabaseTableApplied:   TableName: Exams Provision:All
DatabaseTableApplied:   TableName: PracticeSettings Provision:All
DatabaseTableApplied:   TableName: Scans Provision:All
DatabaseTableApplied:   TableName: AspNetUsers Provision:All
DatabaseTableApplied:   TableName: Locations Provision:All
DatabaseTableApplied:   TableName: Patients Provision:All
DatabaseApplied:        TableName: Practices Provision:All
DatabaseApplied:        Tables count:175 Provision:All
TableChangesSelected:   Practices Inserts:0 Updates:0 Deletes:0 TotalChanges:0
TableChangesSelected:   AspNetUsers Inserts:0 Updates:0 Deletes:0 TotalChanges:0
TableChangesSelected:   Exams Inserts:0 Updates:0 Deletes:0 TotalChanges:0
TableChangesSelected:   Locations Inserts:0 Updates:0 Deletes:0 TotalChanges:0
TableChangesSelected:   Patients Inserts:0 Updates:0 Deletes:0 TotalChanges:0
TableChangesSelected:   PracticeSettings Inserts:0 Updates:0 Deletes:0 TotalChanges:0
TableChangesSelected:   ScanData Inserts:0 Updates:0 Deletes:0 TotalChanges:0
TableChangesSelected:   Scans Inserts:0 Updates:0 Deletes:0 TotalChanges:0
TableChangesApplied:    Practices State:Added Applied:1 Failed:0
TableChangesApplied:    AspNetUsers State:Added Applied:11 Failed:0
Sync Exception: SQLite Error 19: 'FOREIGN KEY constraint failed'.. Type:Data.`
I attached a diagram of the azure db.

`_clientProvider = new SqliteSyncProvider("box.db");
 _proxyClientProvider = new WebProxyClientProvider(new Uri(_configuration["ClaWebApiServer"] + "/sync/post"));
[...]

PR #169, opened 2019-09-17
Fix Issue #158: SQLite Error 19: 'FOREIGN KEY constraint failed'

workgroupengineering:
Issue #158 depends on bad table ordering and the fact that sqlite does not support disabling Constraints.

This PR solves the problem by sorting the tables based on the relationships between them regardless of whether the provider supports or not the disabling of constraints.


Mimetis:
Thx @workgroupengineering for this PR !

Issue #174, opened 2019-12-03
[Enhancement/Question] Differ the sync


MrKniight:
In a classic setup with a central database called the Master, getting requests from different clients :
is there a way  to be able to "differ" the sync (maybe with sync configuration parameters but i don't think) ?
As in : get the last changes from the Master server without uploading the new things from the client to the Master.

The idea behind that is to be able to have a kind of "pending" work on the client, and only commit it when the user decides it, BUT the client still need to get "fresh" datas from the Master regularly. The thing is right now, if we "sync" and there's no conflict, everything will be uploaded and downloaded.

As an added (so 2nd) feature to that : while getting the datas from the Master, if there are conflicts, let the client choose "keep mine" or "take server" but still not commit changes immediately to the Master. Right now it is of course possible to intercept conflicts but that would mean "postpone" them and still choose a resolution everytime we would get datas from the Master.

Thanks

Mimetis:
### 1 - Differing datas.

Actually, it's not possible, with the version `0.3`.
In a future version (probably not the version `0.4`, maybe `0.5`) I'm thinking about let the user (from the client perspective) decides from which point in time he wants to sync its data.
So basically in your case, you would be able to sync T0, T1, T2, T4 from the server in a `DownloadOnly` option, and then make a `Bidirectional` sync from T1 to T5.
It's on the roadmap, but be sure, not today or tomorrow, maybe next year :)

### Conflicts

Be careful, all conflicts are resolved on the server side. 
It's always the server who decices which row is the correct one.
I don't think I will bring a feature for postpone conflicts, because, the databases will be then in a "de-synchronized" state.

MrKniight:
yes dont worry the resolution only in server side was very well explained in the how-to.
So
> So basically in your case, you would be able to sync T0, T1, T2, T4 from the server in a DownloadOnly option, and then make a Bidirectional sync from T1 to T5.
does that mean if the download would find conflicts on client, we have to solve them immediately ? or downloadonly doesnt care about conflicts ?

about point 1 : i forgot to add an other question which is related in some way : 
are we able just with the database to find (for extraction) all the changes related to "one sync operation" ? meaning being able to discover what we call the "T0, T1... T5" changes. I only see that being doable by selecting on "update_timestamp" or "last_change_datetime" (at first i thought scope_id was maybe usable but it's not related to that).

but yeah, those enhancements are surely heavy, no doubt about that.

Mimetis:
[...]

Issue #149, opened 2019-05-02
Question: How can I customize the conversion of data types?


staterium:
Hi, is it possible to change how the sync provider stores data in client DB? I want to receive DateTimeOffset from server, and store as Ticks (int64) on client. Reason is that PCL SqlLite client wrapper stores datetimeoffset as ticks, so any records added on client looks different to sync provider.

Mimetis:
I think about implementing this kind of service, for a next version.   
If you have any idea, we can discuss about it !

Mimetis:
See documentation here https://mimetis.github.io/Dotmim.Sync/SerializerConverter

Issue #110, opened 2018-10-04
[Enhancement] - Callback on client for Downloaded/Conflicting rows


gentledepp:
Think about this scenario:

* On the client side, you may have edited some customer data, which is still open in your view.
* During the sync, the customer record you just edited is rejected by the server (as some guy modified that employee first)
* So Dotmim.Sync sends back the server-record of the customer which is then used to update the clients database.

The problem is:
as there is no callback on the client side, that some conflicting records have been downloaded, we have to reload all user interfaces every time we synchronized with dotmim.sync.
Even if we do so, the user may not realize that all his/her changes have just been overruled by some other data.

=> it would be great to get some "Progress" callbacks on the client side, where the downloaded and (most importantly) conflicting records can be examined.
Then we could fire UI notifications and the like.

The same would be great on the server side where we could modify/accept/reject any records received by a client application.

What do you think?

Mimetis:
You can definitively do this kind of spying system with the last version.
Let me know if you need a piece of code to demonstrate it

gentledepp:
I always like to see a "piece of code" 😁

Mimetis:
Not sure I did mention somewhere that now, in the last version, we have the rows to be applied, but you have a simplified version here : https://mimetis.github.io/Dotmim.Sync/Progression

And here is a more complete version, really verbose in my opinion :D

``` cs
// Database script used for this sample : https://github.com/Mimetis/Dotmim.Sync/blob/master/CreateAdventureWorks.sql 

var serverProvider = new SqlSyncProvider(serverConnectionString);
var clientProvider = new SqlSyncProvider(clientConnectionString);
var tables = new string[] {"ProductCategory", "ProductModel", "Product",
            "Address", "Customer", "CustomerAddress", "SalesOrderHeader", "SalesOrderDetail" };

// Creating an agent that will handle all the process
var agent = new SyncAgent(clientProvider, serverProvider, tables);

// First sync to initialize everything
var s1 = await agent.SynchronizeAsync();
// Write results
Console.WriteLine(s1);

// Make some changes on the server side
var sc = serverProvider.CreateConnection();
await Helper.InsertOneCustomerAsync(sc, "John", "Doe");
await Helper.InsertOneProductCategoryAsync(sc, "Shoes 2020");
await Helper.DeleteOneSalesDetailOrderAsync(sc, 113141);
await Helper.DeleteOneSalesDetailOrderAsync(sc, 113142);
[...]

PR #177, opened 2019-12-27
V 0.4

Mimetis:
This PR is the next version of Dotmim.Sync.
It includes a lot of new features and improvements.

This version **IS NOT** compatible with the actual version (`0.3`)

> **Update**: The `v0.4` packages are available as *pre-release* on nuget : https://www.nuget.org/packages?q=dotmim.sync

Here are some new features implemented:

## Change Tracking

New **Sql** provider that uses the **Change Tracking** feature, from SQL Server.
This provider is called `SqlSyncChangeTrackingProvider`.

`SqlSyncChangeTrackingProvider` is compatible with all others sync providers.

What does it mean ?
- No more tracking tables in your database
- No more triggers on your tables in your database
- Metadatas retention managed by SQL Server itself 
- Some restrictions on filters (especially on Delete statement with a filter)

To be able to use `SqlSyncChangeTrackingProvider` on your database, do not forget to active the **Change Tracking** on your database :

``` SQL
ALTER DATABASE AdventureWorks  
SET CHANGE_TRACKING = ON  
(CHANGE_RETENTION = 14 DAYS, AUTO_CLEANUP = ON)  
```

You don't have to activate **Change Tracking** on each table. It will be enabled by the `Dotmim.Sync` framework.

## Tracking tables

If you're not using the `SqlSyncChangeTrackingProvider` provider, your database will contain some tracking tables.

Un-necessary columns are removed from `_tracking` tables
- `create_scope_id` : Removed
- `create_timestamp` : Removed
- `update_timestamp` : Removed
- `update_scope_id` : Maintained : We need to know *who* is uploading data to a datastore, because the datastore should not select these data when he makes the `SelectChanges` method

Using a table `Address` as example, the `_tracking` table structure is now:

``` SQL
CREATE TABLE [dbo].[Address_tracking](
	[AddressID] [int] NOT NULL,
	[update_scope_id] [uniqueidentifier] NULL,
	[timestamp] [timestamp] NULL,
	[timestamp_bigint]  AS (CONVERT([bigint],[timestamp])) PERSISTED,
	[sync_row_is_tombstone] [bit] NOT NULL,
	[last_change_datetime] [datetime] NULL,
```
> `timestamp_bigint` is a compute column, only created on **SQL Server** to allow a better index on the `_tracking` table.

A new index is creating as well on `_tracking` table:

```
CREATE NONCLUSTERED INDEX [Address_tracking_timestamp_index] ON Address_tracking
([timestamp_bigint] ASC
,[update_scope_id] ASC
,[sync_row_is_tombstone] ASC)
```

## Server scopeinfo table

The `sync_scope` table disappears from the server side data source.
This **PR** changes the role regarding the management of the `LastSyncTimestamp`. 

Now, the client is responsible of storing the values from both **Server** side and **Local** side:

``` sql
SELECT TOP (1000) [sync_scope_id]
      ,[sync_scope_name]
      ,[sync_scope_schema]
      ,[scope_last_server_sync_timestamp]
      ,[scope_last_sync_timestamp]
      ,[scope_last_sync_duration]
      ,[scope_last_sync]
  FROM [Client].[dbo].[scopeinfo]
```

sync_scope_id | sync_scope_name | sync_scope_schema | scope_last_server_sync_timestamp | scope_last_sync_timestamp | scope_last_sync_duration | scope_last_sync
-------------|---------------------|--------------------|-------------|---------------------|--------------------|--------------------
F6780CB1... | DefaultScope  | { }   |  10001  | 7815753  |  2019-12-27 06:12:15.617

With this new feature, the server is completely independant, and is able to response to any requests, from any clients, and any timestamps requested.

## Less Roundtripts

Today we have at least 6 requests-responses roundtrips:

> 1) BeginSession
> 2) EnsureScopesAsync
> 3) EnsureSchemaAsync
> 4) EnsureDatabaseAsync
> 5) ApplyChangesAsync
> 6) GetLocalTimestampAsync
> 7) GetChangeBatchAsync
> 8) WriteScopesAsync
> 9) EndSessionAsync

It's quite a lot of messages, and so far, a lot of opportunities to fail for various reasons (network and so on)   
Almost all steps has its own transaction, and it could be potentially not safe, if all the steps are not correctly done.

This **PR**  will introduce a new concept of `Orchestrator`.   
An orchestrator is responsible of the communication and need a provider (which is responsible of the datastore)   

At the end, we have now only 2 requests-response:

> 1) EnsureScopeAsync
> 2) ApplyThenGetChangesAsync

Basically, we have **2 transactions** on the client side and only **1 transaction** on the server side.

This PR will include a new **Schema** cache, on the client side.
Once the schema readed from the server, the client will store the schema in its `sync_scopes` table.
So far, we reduce the roundtrip to **1** query:

> 1) ApplyThenGetChangesAsync

Here is a screenshot, comparing `v0.3` and `v0.4`:

image

## Replacing DmSet with SyncSet

`DmSet` is a great object, that simulate a `DataSet`.  
I have created thie object, because at the beginning, `.Net Core 1` did not have the `DataSet`.

Today the `DmSet` is too complicated to handle, especially when optimizing **memory** and **network**

The `SyncSet` object is able to serialize its schema without the datas, and is able to export the datas in a convenient way (using an object called `ContainerSet`)

Here is a comparison, regarding the serialization of the schema :

- **V0.3** : Schema serialization : **45 961 bytes**
- **V0.4** : Schema serialization : **11 580 bytes** 

## Client decides which serializer to use

Client can dictate the serialization mechanism
The client sends a special HTTP header `dotmim-sync-serialization-format`, containing two information:

- First one is specifying the serialization format to use. The server then knows how to deserialize the messages and also uses this serialization format in each response
- Second one is specifying if the client needs batch mode or not.

Example:

```
dotmim-sync-serialization-format: {"f":"json","s":500}
```
- Serialization : **Json**
- Batch size : **500**

## Options to set a custom serializer

You can now set your own serializer. 
Using the `SyncOptions` instance, the client can define the serializer he wants to use.

The server shoud have the serializer installed as well.

To be able to use a new serializer, you should:
- Implements thes interfaces `ISerializerFactory` and `ISerializer<T>`
- References this serializer on both side (client and server)

Here is an example using a new serializer based on **MessagePack**:

``` csharp
public class CustomMessagePackSerializerFactory : ISerializerFactory
{
    public string Key => "mpack";
    public ISerializer<T> GetSerializer<T>() => new CustomMessagePackSerializer<T>();
}

public class CustomMessagePackSerializer<T> : ISerializer<T>
{
    public CustomMessagePackSerializer() => MessagePackSerializer.SetDefaultResolver(MessagePack.Resolvers.ContractlessStandardResolver.Instance);
    public T Deserialize(Stream ms) => MessagePackSerializer.Deserialize<T>(ms);
    public byte[] Serialize(T obj) => MessagePackSerializer.Serialize(obj);
}
```
Add the serializer to the web server serializers collection:
``` csharp
var webServerOptions = new WebServerOptions
{
    UseBulkOperations = ...
};
webServerOptions.Serializers.Add(new CustomMessagePackSerializerFactory());

```

Add the serializer on the client side:
``` csharp
var proxyClientProvider = new WebClientOrchestrator
{
    SerializerFactory = new CustomMessagePackSerializerFactory()
};
```

## Options to set a custom converter

Allows you to convert each row **before being serialized** (when send a request) and **after being deserialized** (when receiving a response)

Like the `ISerializerFactory`, you can implement your own `IConverter`.
- This convert should be available both on the client and the server.
- The server registers all converters used by any client
- The client register its own converter.

``` csharp
public interface IConverter
{

    /// <summary>
    /// get the unique key for this converter
    /// </summary>
    string Key { get; }

    /// <summary>
    /// Convert a row before being serialized
    /// </summary>
    void BeforeSerialize(SyncRow row);

    /// <summary>
    /// Convert a row afeter being deserialized
    /// </summary>
    void AfterDeserialized(SyncRow row);
}
```

Example of a simple `IConverter`:

``` csharp
    public class CustomConverter : IConverter
    {
        public string Key => "cuscom";

        public void BeforeSerialize(SyncRow row)
        {
            // Each row belongs to a Table with its own Schema
            // Easy to filter if needed
            if (row.Table.TableName != "Product")
                return;

            // Encode a specific column, named "ThumbNailPhoto"
            if (row["ThumbNailPhoto"] != null)
                row["ThumbNailPhoto"] = Convert.ToBase64String((byte[])row["ThumbNailPhoto"]);

            // Convert all DateTime columns to ticks
            foreach (var col in row.Table.Columns.Where(c => c.GetDataType() == typeof(DateTime)))
            {
                if (row[col.ColumnName] != null)
                    row[col.ColumnName] = ((DateTime)row[col.ColumnName]).Ticks;
            }
        }

        public void AfterDeserialized(SyncRow row)
        {
            // Only convert for table Product
            if (row.Table.TableName != "Product")
                return;

            // Decode photo
            row["ThumbNailPhoto"] = Convert.FromBase64String((string)row["ThumbNailPhoto"]);

            // Convert all DateTime back from ticks
            foreach (var col in row.Table.Columns.Where(c => c.GetDataType() == typeof(DateTime)))
            {
                if (row[col.ColumnName] != null)
                    row[col.ColumnName] = new DateTime(Convert.ToInt64(row[col.ColumnName]));
            }
        }
    }
```

On client side, register this converter from your `WebClientOrchestrator`:

``` csharp
// Create the web proxy client provider with specific options
var proxyClientProvider = new WebClientOrchestrator
{
    SerializerFactory = new CustomMessagePackSerializerFactory(),
    Converter = new CustomConverter()
};

```

On server side, add this converter to the list of available converters:

``` csharp
var webServerOptions = new WebServerOptions
{
   ...
};
webServerOptions.Serializers.Add(new CustomMessagePackSerializerFactory());
webServerOptions.Converters.Add(new CustomConverter());

```

Without Converter:

image

With Converter:

image

### 1 - HttpClient
To be able to use the last features from **ASP.NET Core**, especially Dependency Injection of `HttpClient`, the `WebClientOrchestrator` is taking an `HttpClient` as parameter

``` csharp
public WebClientOrchestrator(string serviceUri = null, ISerializerFactory serializerFactory = null, HttpClient client = null)
{
}
```

### 2 - Gzip compression

Gzip compression can be used, but is not enabled by default. Use an instance of `HttpClientHandler` and check the `SupportsAutomaticDecompression` property:

``` csharp
var handler = new HttpClientHandler();

if (handler.SupportsAutomaticDecompression)
    handler.AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate;

var httpClient = new HttpClient(handler);

var client = new WebClientOrchestrator("http://...", SerializersCollection.JsonSerializer, httpClient);

```


gentledepp:
### 1 Nuget errors
I tried to build this branch and for some reason (although appveyour seemingly shows the green check mark), I do have 2 build errors:

image

### 2 Build errors
When fixing those nuget errors, it turns out that dotnet-sync does not build.
image

Did you omit this by purpose?


### 3 Failing tests
Also, I do have many failing tests on my machine:

Mimetis:
Hey @gentledepp 

## Nuget errors

I've updated all the nuget packages for all projects. Maybe you should clean your solutions and try to build again everything.
In your screenshot, I see errors from `dotnet-sync` and `SampleWebServer` 
- `dotnet-sync` : this project is not up to date. I think I will mark this project as **Obsolete** until I reach a really stable version on the core projects.
- `SampleWebServer`: I will update the sample asap. Just unload it.

## Build Errors

In my solution, I've only these projects loaded:
- `Dotmim.Sync.Core`
- `Dotmim.Sync.SqlServer`
- `Dotmim.Sync.MySql`
- `Dotmim.Sync.Sqlite`
- `Dotmim.Sync.Web.Client`
- `Dotmim.Sync.Web.Server`
- `Dotmim.Sync.Tests`
- `Dotmim.Sync.SampleConsole`

## Failing tests

I will update the `Standalone` tests asap. `DmSet` is not used anymore in `Dotmim.Sync`. Anyway the tests must be green :)

For `SerializerTests`, I will check asap.

Once again, the most important tests are `SqlServer` `Sqlite` and `MySql`: If they are green, it's a good point :)

## Devops

Be careful, you used to say **appveyor** but it's actually **Azure devops** :)
On the devops pipeline, the StandAlone tests are not included.

See the def in `:/azure-pipelines.yml`:

``` yml
     - script: dotnet test Tests/Dotmim.Sync.Tests/Dotmim.Sync.Tests.csproj --filter SqlServerBasicTests --logger trx
        name: SqlServerBasicTests
        displayName: "Tests with SQL Server acting as server"

      - script: dotnet test Tests/Dotmim.Sync.Tests/Dotmim.Sync.Tests.csproj --filter SyncAllColumnsTests --logger trx
        name: SyncAllColumnsTests
        displayName: "Tests with SQL Server acting as server and all columns possibilities"

      - script: dotnet test Tests/Dotmim.Sync.Tests/Dotmim.Sync.Tests.csproj --filter SqlServerFilterTests --logger trx
        name: SqlServerFilterTests
        displayName: "Tests with SQL Server acting as server and specific filters"

      - script: dotnet test Tests/Dotmim.Sync.Tests/Dotmim.Sync.Tests.csproj --filter SqliteEncryptedTests --logger trx
        name: SqliteEncryptedTests
        displayName: "Tests with SQL Server acting as server and encrypted SQLite database"

      - script: dotnet test Tests/Dotmim.Sync.Tests/Dotmim.Sync.Tests.csproj --filter MySqlBasicTests --logger trx
        name: MySqlBasicTests
        displayName: "Tests with MySQL Server acting as server"

      - script: dotnet test Tests/Dotmim.Sync.Tests/Dotmim.Sync.Tests.csproj --filter MySqlFilterTests --logger trx
        name: MySqlFilterTests
        displayName: "Tests with MySQL Server acting as server"

```

gentledepp:
Ok! Thanks for the heads-up!
I already thought that my dev machine had some serious problems as you mentioned that 
> all tests pass
and on my end they did not.

I will continue to analyze your solution.
My prominent focus will be backward compatibility, because:
 - We have many v0.3 clients out there in the wild and also v0.3 servers
 - We want to make sure that our servers are compatible to v0.3 **and** v0.4 for some time (e.g. 1 year) until we are sure that all clients have successfully upgraded to v0.4 or later.

In the worst case, I will have to create
- a private package of dotmim.sync v0.3**.1** which changes the namespaces so that they do not conflict with v0.4 and migrate our current server to run v0.3.1 and v.04 in parallel, so old and new clients can be served with their preferred protocol

This is going to be a lot of work I assume, but let's see

Mimetis:
Don't go to fast on migration ! 
I still have a lot of features I want to implement before marking `Dotmim.Sync` as **v1.0**

Until then, just stay on `v0.3` if it's working well for you.
The only reason you should migrate to `v0.4` or +, is because the performances are too low for you on `v0.3`

Otherwise, stay on `v0.3` and look at `v0.4` improvements and give me feedbacks :)

My next feature (and the new architecture allows me to implement it, quite easily) will be **Snapshot** initialization batch
- Server creates a *snapshot* of the database at time T1 (let's say, every 15 days ?)
- Each **new** client will download this snapshot for initialization
 - client will apply this snapshot
 - client will then sync from this T1 time to T0

The idea is to have a light initialization, especially on big databases. Today each client requests its own batch when initialize. It's a lot of works for the server, and sometimes can lead to timeout.

Since you're using the `Dotmim.Sync` fx in production (the only one in the world I guess ;) ), how do you handle this particular phase (init db on client) ?

As I said, too many things to do :D

Mimetis:
Tests fixed !

gentledepp:
Well, we actually **do** have to go fast on migration, as the current implementation of sync (v0.3) is the main pain point for our clients. We are spending hours fixing sync issues that should never occur.
So waiting and sticking to v0.3 is not an option for us, unfortunately.

The feature **snapshot initialization batch** is not really relevant to us.
The simple reasons are
1. We do not have to optimize on first time/full synchronization so much as this does not happen so often. (1000 users => maybe 2000 times a year, but we do have way more incremental syncs)
2. As our clients run on mobile phones, they are quite resource-constrained, which is why we try to sync as little data as possible and just as much as necessary. So basically, each device sends along the current `userId` and that user gets its very own personalized set of data.
This means potentially 1000 snapshots for 1000 uses which would not make any sense obviously

I would argue, that more important implementations are
1. Improving the efficiency of a sync process (by caching as much as possible - e.g. the DbCommands as I described in the issue, using SQL indices, etc.)
2. Specifying that only a subset of columns is synced
3. (Maybe) sync filters
We do already have a very good solution for this, though ;-)

gentledepp:
> Tests fixed !
Thanks, looking quite good now.
Every 2nd run, all tests pass.
But all the other times, there seems to be one test `SqlServerBasicTests` that fails causing all remaining tests to fail as well.

This would be another important "feature" - at least for contributors to Dotmim.Sync
Currently there only exist integration tests, no unit tests.
While integration tests are great, sometimes they make it more cumersome to to test very specific behabvior.
And even worse: The test methods build up on each other and make them behave like domino stones: Once one falls, all others follow suit, which makes it very hard to track down errors sometimes.
It would be **MUCH** better if each test would be completely independent, meaning I can run any test at any time separately and it will still succeed

Why did you choose to go down this domino approach?

Mimetis:
> as the current implementation of sync (v0.3) is the main pain point for our clients. We are spending hours fixing sync issues that should never occur.

Really sorry about that :(
Please submit issues. Will try to help you as much as possible.
Unfortunatelly, this project is a *side project* and sometimes, I lakc times to investigate more :(

> Currently there only exist integration tests, no unit tests.

Indeed, we should have unit tests. Lot of work, I guess...

> The test methods build up on each other and make them behave like domino stones
> Why did you choose to go down this domino approach?

Ah ah, I don't know :) 
For sure, I should simplify things. Don't worry, it's also very annoying for me: D

Mimetis:
As you proposed in one of your last comments, I've create a new stored procedure called `[table]_initialize` that serves for:
- First sync on a new client db
- Sync when method `Synchronize` is called with `Reinitialize` or `ReinitializeWithUpload`

With this addition, the `[table]_SelectChanges` stored procedure is way more effective, with less parameters and a very simplified query, with only 2 parameters:

If we take a table called `ProductCategory`, here are the `Where` query part :

 `ProductCategory_SelecChanges` without a filter:

``` sql
SELECT ....
FROM [ProductCategory] [base]
RIGHT JOIN [tProductCategory] [side]
ON [base].[ProductCategoryID] = [side].[ProductCategoryID]
WHERE (
	[side].[timestamp] > @sync_min_timestamp
	AND ([side].[update_scope_id] <> @sync_scope_id OR [side].[update_scope_id] IS NULL)
)

```

 `ProductCategory_SelecChanges` with a filter:

``` sql
SELECT ....
FROM [ProductCategory] [base]
RIGHT JOIN [tProductCategory] [side]
ON [base].[ProductCategoryID] = [side].[ProductCategoryID]
WHERE (
	([side].[ParentProductCategoryID] = @ParentProductCategoryID )
	OR ([side].[ParentProductCategoryID] IS NULL )
	)
AND (
	[side].[timestamp] > @sync_min_timestamp
	AND ([side].[update_scope_id] <> @sync_scope_id OR [side].[update_scope_id] IS NULL)
)

```

 `ProductCategory_Initialized` without a filter:
``` sql
SELECT ...
FROM [ProductCategory] [base]
-- Hum ... that's all :)
```

`ProductCategory_Initialized` with a filter:

``` sql
SELECT ...
FROM [ProductCategory] [base]

WHERE 
	([base].[ParentProductCategoryID] = @ParentProductCategoryID )
 OR ([base].[ParentProductCategoryID] IS NULL )
```

gentledepp:
Nice!

One thing to note though: I see you are introducing a lot of breaking changes, like this one with new stored procedures.

So please
 - [ ] Either write down in a readme all steps to take to migrate from 0.3 to 0.4, so you can create a "how-to" later
 - [ ] Or (much better) carefully craft those changes with sensible defaults, so that no changes are required by developers to migrate.

In the case of the stored procedures, it would e.g. be good if
- the old "select_changes" stored procedure has default values so that they will still work as `SelectChangesWitFilters` with version 0.4 and you do not have to overwrite it
- and the setup/init logic of dotmim.sync that sets up the tracking tables and stored procedures makes sure that it also checks for the existence of those new procedures (initialize) even if all other stuff is already in place so that you can simply run it on a 0.3 database schema and it will still work (and automagically 🧙‍♂️ migrate to a 0.4 schema)


- or - much easier - just change the name of the `sp_selectchanges` so that the old one from version 0.3 still exists

Because some people out there might have to migrate dotmim.sync fom 0.3 to 0.4 you know... 😅
(And please forgive me if you already take care of these matters and I did not see it)

Mimetis:
Wouldn't it be easier to just make a `DeprovisionAsync` and then a `ProvisionAsync` again, when you try to migrate ?

I mean, when deprovisioning, you can keep the tracking tables in place, if you specify the rights params to `DeprovisionAsync()`
That way, you just delete all stored proc / triggers, then regenerate them again.

Should be easier ?

Mimetis:
By the way, just introduced a new way to specify columns when initialize : `SyncSetup`

Quick sample:

``` csharp
        var setup = new SyncSetup(new string[] { "Product", "ProductCategory"});
        setup.Tables["Product"].Columns.AddRange(new string[] { "ProductId", "Name", "ProductCategoryID", "ProductNumber", "StandardCost", "ListPrice", "SellStartDate", "rowguid", "ModifiedDate" });

        // Add pref suf
        setup.StoredProceduresPrefix = "s";
        setup.StoredProceduresSuffix = "";
        setup.TrackingTablesPrefix = "t";
        setup.TrackingTablesSuffix = "";

```

gentledepp:
> By the way, just introduced a new way to specify columns when initialize : SyncSetup
😍😍😍😍😍😍😍!!!! I'm loving it!

> Wouldn't it be easier to just make a DeprovisionAsync and then a ProvisionAsync again, when you try to migrate ?
Well... yes, you could keep it that way. However, you'd have to inform all developers of the fact, that they need to re-provision.

For **us** this is not an option, because we are already using 0.3 in production and cannot do any "big bang" rollout anymore. 
Remember: We do have about 1000 clients out there running 0.3, so we need to run 0.3 and 0.4 side-by-side on the server side for some time.
If 0.4 was still backwards compatible, this would not be necessary, but let's be honest: It is not ;-)

Mimetis:
Yes, it's not compatible.
As I mentionned somewhere, until we reach the `v1.0`, we could potentially introduce a lot of breaking changes.
Anyway, I will try to find a way to migrate from `v0.3` to `v0.4` with the less problems as possible.

gentledepp:
Well, our current plan is to
- create a build of 0.3 for the server side (0.3.1), where we just change the namespaces, so we can run our 0.3.1 version in parallel to 0.4

for this to work, however, the database scheme must be the same or at least compatible.
This is why I am so concerned with the change to the `sp_selectchanges` as we do not want to introduce yet another scope to our server database.
- we have the v1 scope with v1 tracking tables still present from Microsoft Sync Framework
- we have the v2 scope with v2 tracking tables running dotmim.sync 0.3
- we **do not want** to introduce a scope v3 for dotmim.sync 1.0

So in the worst case, we'll have to "hard code" the name of the `sp_selectchanges` of our own 0.3.1 version, so that it is not in the way of 0.4

Mimetis:
Regarding naming conventions, is it a reasonable solution to just add **suffix** and **preffix** to `v.04` for stored procedures ? 
I mean, something like:

``` csharp
        var setup = new SyncSetup(allTables);

        // Add pref suf for v0.4
        setup.StoredProceduresPrefix = "vfour";
        setup.StoredProceduresSuffix = "";
        // for tracking tables and suffix, we still want the same
        // so replace with same values as v0.3
        setup.TrackingTablesPrefix = "";
        setup.TrackingTablesSuffix = "";
        setup.TriggersPrefix = "";
        setup.TriggersSuffix = "";

```
Anyway, I will change the **SelectChanges** stored procedure name, if it can help you
Go for `sp_changes` instead of `sp_selectchanges` ?

## Do you have any plans for when 1.0 will be released?

No idea !

gentledepp:
> Regarding naming conventions, is it a reasonable solution to just add suffix and preffix to v.04 for stored procedures ?

You are right! If we can setup a scope configuration where we can just modify the sp-names, then that should be perfectly fine for us! 😁🎉

> No idea !

Do you at least have in mind a list of features that is **absolutely required** to make for a 1.0 release?

In my opinion, with the addition of the possibility to specify which columns to sync (again, thank you so much - I am loving it!), Dotmim.Sync seems to be feature complete for a stable release.
This is because it supports everything Microsoft Sync Framework does and more.

Other features are rather a nice-to-have and do not necessarily change the functionality that  much
1. snapshots for inizialization (again, for us this is not usable anyways)
2. specify different batch size for uploads and downloads (nice to have)
3. advanced sync filters (nice to have, but can be added as an additional module. I am afraid that your original attempt to solve this with support for cross-table-filters will not ever work, as it is too complex to implement and also too slow from a querying perspective)
We did implement our own sync filter "framework" by modifying the select_changes stored procedures to include our hand-crafted filters - thanks to the interceptors ;-))
4. Having interceptors also get a hold of all the rows that are about to be inserted/updated locally (or after they were inserted) (again: nice to have and can be also implemented differently: e.g. querying for any changes since the last sync timestamp)
5. performance improvements (in-memory schema caching, dbcommand caching, memory management, etc. - non-breaking changes that can be added on-the-go)

and honestly, I currently cannot come up with any more items for this "whishlist" :-)
Is there anything else on your mind?

Mimetis:
## 3. advanced sync filters

Really interested in your solution. Let's discuss it here #179

## 4. Having interceptors also get a hold of all the rows that are about to be inserted/updated of after they were inserted/updated locally

Seems to be easy, can be done quickly in my opinion.

## 5. performance improvements

Strings allocation have been greatly improved in `v0.4`, especially because we don't use `DmSet` anymore and all rows are allocated in one single `object[]` array.
I still did not do any performance test yet, but I'm pretty sure we gain a lot.

Regarding dbCommand caching, do you have any pointer on why this is so important ?

In memory schema caching : Seems done in `v0.4`

## 6. customize conversion of data types

Yes, this feature is a must have. 
But complicated due to the differents serializers we are using (json, datacontract, binary, messagepack)

*I'm still struggle with MessagePack since the conversion of a `byte[]` array (stored in an `object[]`) does not work properly yet.*

## 7.  Conflict resolution on client side #174 and related #110

You can set the `ConflictResolutionPolicy` property (in `SyncOptions`) to `ConflictResolutionPolicy.ClientWins`.
That way, the conflict will be resolved on the server side, but the client row will win the conflict and will override the server row.

Maybe you want also to be able to use the `Merge` action, but not from server but from client side.  
This is more complicated since it's always the server who eventually resolve the conflict.

Anyway, theoricaly, it's faisable, but I don't think it will be something available for `v1.0`

## 9. Deleted rows remain in _tracking table

Yes, it's a must have. The `CleanMetadata` options has been created for this, but never implemented :)

## 8. 10. 11. 12.
Done

## 13. Additional providers

Later :p

## 🥇 Mandatory features for stable release

In my opinion, we should be able to go to `v1.0` if we resolve:

- Bug fixes
- Sync filters
- Implements `CleanMetadatas`

gentledepp:
### 3. advanced sync filters

Well this is a complex topic and hard to explain in writing on github.
If you like, we can have a skype chat some time and I will try to prepare something and to explain.

Basically, our concept visible in a repository on github:
https://github.com/gentledepp/SyncFilters


### 5. performance improvements

>Regarding dbCommand caching, do you have any pointer on why this is so important ?

This is because whenever you send a commend to e.g. SQLite and MSSQL server, the SQL has to be parsed, and an execution plan determined.
If you cache the DbCommand, this is not necessary anymore.
*Your* SQLite port of Microsoft Sync Framework already used this caching mechanic!
This is why I am a little puzzled that you seem to not know anything about it 😅

> In memory schema caching : Seems done in v0.4

ok, so how can I make sure that the schema is cached in-memory?
- for the lifetime of asp.net applicationpool (singleton)
- for the lifetime of my android app (singleton)

One point missing in the list of performance improvements is the caching logic we added for `ObjectNameParser` as for larger data sets the repetitive parsing led to bad performance. Parsing always is expensive!

### 6. customize conversion of data types
Well, this should not have anything to with the serializers, but should be 
- run, before a message is serialized and sent
- and run after a message was received and deserialized

For MsgPack, we are using MessagePack-CSharp and one of our problems was, that it its highly optimized: It created IL code on-the-fly, but that does not work on iOS for example.
So we have to generate the serialization logic beforehand
https://github.com/neuecc/MessagePack-CSharp#aot-code-generation-to-support-unityxamarin

Also, you can add your own resolvers for specific types.
- there is already a DateTime resolver, so when you serialize a DateTime as object and deserialize, you *will* actually get a DateTime back
- for Guids, this does not work and gives you a string!

Long story short: Using messagepack *is* achievable, but out of scope of Dotmim.Sync v1 I guess.

### 7. Conflict resolution on client side #174 and related #110
Well, I am aware of `ClientWins`, but that is something completely different.
What the guy from #174 wanted is to actually solve conflicts on the client side.

This *could* be implemented, by *not* sending all changes to the server, but wait for the date from the server first.
Then, detect any conflicts locally and fix them, and finally do another *upload-only-sync*

This obviously does not work very well with your new "simplified-flow" protocol.
And yes, I agree that this is not a must-have for v1.


## 🥇 Mandatory features for stable release
In my opinion, we should be able to go to v1.0 if we resolve:

- [ ] Bug fixes **absolutely*
- [ ] Sync filters **...hmmm...** well I fear your solution will never work, and ours is conceptually an addition **on top of** dotmim.sync, because we use application logic to determine what to sync and what not. 
- [ ] Implements CleanMetadatas **absolutely**
This seems to be a trivial task, or am I missing something here?

gentledepp:
Small side-note: Since your last change (configurable columns to sync), 15 tests fail constantly.

For whatever reason, the SqlServerBasicTests fail, because they cannot connect to **MySql** ⁉ 😂
image

Mimetis:
Yes, just because you don't have mysql locally (or at least, you don't have the same local config)
Go check this file `/Tests/SqlServer/SqlServerBasicTests` and replace this line

``` csharp
if (!Setup.IsOnAzureDev)
{
    providerFixture.AddRun(NetworkType.Http,  ProviderType.MySql);
    providerFixture.DeleteAllDatabasesOnDispose = false;
}
```

by this:

``` csharp
if (!Setup.IsOnAzureDev)
{
    providerFixture.AddRun(NetworkType.Http,  ProviderType.Sql);
    providerFixture.DeleteAllDatabasesOnDispose = false;
}
```

Basically, it says 
- *"Make a run on `HTPP`, using a client database of type `Sql`"*
- *Don't delete client databases at the end*

Mimetis:
Just added a new feature : `IConverter`:

Allows you to convert each row **before being serialized** (when send a request) and **after being deserialized** (when receiving a response)

Like the `ISerializerFactory`, you can implement your own `IConverter`.
- This convert should be available both on the client and the server.
- The server registers all converters used by any client
- The client register its own converter.

``` csharp
public interface IConverter
{

    /// <summary>
    /// get the unique key for this converter
    /// </summary>
    string Key { get; }

    /// <summary>
    /// Convert a row before being serialized
    /// </summary>
    void BeforeSerialize(SyncRow row);

    /// <summary>
    /// Convert a row afeter being deserialized
    /// </summary>
    void AfterDeserialized(SyncRow row);
}
```


Example of a simple `IConverter`:

``` csharp
    public class CustomConverter : IConverter
    {
        public string Key => "cuscom";

        public void BeforeSerialize(SyncRow row)
        {
            // Each row belongs to a Table with its own Schema
            // Easy to filter if needed
            if (row.Table.TableName != "Product")
                return;

            // Encode a specific column, named "ThumbNailPhoto"
            if (row["ThumbNailPhoto"] != null)
                row["ThumbNailPhoto"] = Convert.ToBase64String((byte[])row["ThumbNailPhoto"]);

            // Convert all DateTime columns to ticks
            foreach (var col in row.Table.Columns.Where(c => c.GetDataType() == typeof(DateTime)))
            {
                if (row[col.ColumnName] != null)
                    row[col.ColumnName] = ((DateTime)row[col.ColumnName]).Ticks;
            }
        }

        public void AfterDeserialized(SyncRow row)
        {
            // Only convert for table Product
            if (row.Table.TableName != "Product")
                return;

            // Decode photo
            row["ThumbNailPhoto"] = Convert.FromBase64String((string)row["ThumbNailPhoto"]);

            // Convert all DateTime back from ticks
            foreach (var col in row.Table.Columns.Where(c => c.GetDataType() == typeof(DateTime)))
            {
                if (row[col.ColumnName] != null)
                    row[col.ColumnName] = new DateTime(Convert.ToInt64(row[col.ColumnName]));
            }
        }
    }
```

On client side, register this converter from your `WebClientOrchestrator`:

``` csharp
// Create the web proxy client provider with specific options
var proxyClientProvider = new WebClientOrchestrator
{
    SerializerFactory = new CustomMessagePackSerializerFactory(),
    Converter = new CustomConverter()
};

```

On server side, add this converter to the list of available converters:

``` csharp
var webServerOptions = new WebServerOptions
{
   ...
};
webServerOptions.Serializers.Add(new CustomMessagePackSerializerFactory());
webServerOptions.Converters.Add(new CustomConverter());

```

Without Converter:

image

With Converter:

image

Mimetis:
## DbCommands

> This is because whenever you send a commend to e.g. SQLite and MSSQL server, the SQL has to be parsed, and an execution plan determined.
> If you cache the DbCommand, this is not necessary anymore.

Well, SQL Server caches the execution plan, and will reuse it if your resend the same command (with differents parameters)
So, I really don't see where we can gain performance.

In the other hand, for SQLite, it could be different. I don't know how the caching system is working for SQLite.

Did you notice a real performance gain by caching commands for SQLite ?

> Your SQLite port of Microsoft Sync Framework already used this caching mechanic!

I don't remember it :D

gentledepp:
### DbCommands

I cannot remember exactly, but command caching led to significant performance improvements when used in SQLite *and* MSSQL.
MSSQL **may** cache the execution plan, **but** it still has to **parse** the SQL statement each time you create a new DbCommand.
So when you insert 1000 new rows (not using bulk insert), the statement has to be parsed 1000 times.
So this is especially beneficial for database providers, that do not have bulk support - rather MySql than MSSQL.

I'll implement our command caching logic again as soon as this PR is through so that I can create a PR for this to show you ;-)



### IConverter

This is really awesome!
You may want to copy-paste this information into issue #149 
;-)

gentledepp:
### DbCommands

So I just re-ran our performance tests for SQLite to MSSQL Server synchronization.
It syncs just one table with 250.000 or more rows.

We implemented command caching in the DbSyncAdapter, so it is basically automatically implemented for **every platform**.
See our commit here: https://github.com/Opti-Q/Dotmim.Sync/commit/e67d7f38a51db6e2505910f5163120635b42a532

Please just note, that this commit also included some caching of writable columns which has nothing to do with command caching. 

The results are quite interesting when syncing 250.000 rows:
command cache **ON**:  00:00:10.9776233
command cache **OFF**: 00:00:12.4838710

When raising this to 500.000 rows things become quite ugly though:
command cache **ON**:  00:00:14.8558512
command cache **OFF**: 00:00:23.9173400

Notice, that we ran this on a Windows 10 machine with **full .net**.
When being on **Xamarin.Android** the it would be **signifficantly** worse, because the Mono garbage collector is much weaker (performs worse) than the .NET garbage collector.

That is why we App developers have to keep a close eye on memory consumption. 
**Rule of thumb:** *kilobytes hurt, megabytes kill*

So we also created tests using dotMemory Unit.
Note that our tests starts an initial sync first so that all caches are set-up, before measuring the actual memory allocation in a second sync.
So this is not a "cold sync".
Also, to be fair, as the memory profiling runs in an integration test, the memory of **client and server combined** is tracked.
So the following test results basically show the memory allocation on client *and* server.

Memory usage when syncing 50000 rows:
command cache **ON**:  ObjectsCount: 1.436.214; SizeInBytes: 100.715.966
command cache **OFF**: ObjectsCount: 1.770.264; SizeInBytes: 121.549.244

That is already a 21 **MB** difference!

Memory usage when syncing 50000 rows:
command cache **ON**:  ObjectsCount: 7.069.023; SizeInBytes: 496.058.369
command cache **OFF**: ObjectsCount: 8.726.850; SizeInBytes: 599.589.010

This is now over a 100 **MB** difference in allocated memory. (1/6 less allocation)


Still not convinced? 😏

As you see in [our commit](https://github.com/Opti-Q/Dotmim.Sync/commit/e67d7f38a51db6e2505910f5163120635b42a532) it is not a lot of effort to implement.
I'll happily contirbute it once your PR is through.

lordofffarm:
Hey, guys, appreciate the work you're doing, and looking forward for v0.4 and sync filters :)

Mimetis:
I've refactored all the tests.
They are all independent, with no order specified.   
Each test can be run separately without any dependence on any other test.

For simplicity, I've refactored all the test mechanism. 
Now the tests are a little bit longer to write, but way more easier to read.

Updated `SqliteProvider` to be able to read database structure.
Fix some `Pragma` problematic ... (Yes `Pragma` directives should be set **before** opening a transaction...)

Fix almost thousand of bugs...

gentledepp:
Hello!
The list of our intended contributions gets even longer.

### 14. Specify columns to sync per table

> By the way, just introduced a new way to specify columns when initialize : SyncSetup

We already implemented this in some hacky way, but it works.
What I completely forgot was one important detail:

The `update_trigger` *should* ignore changes on a row that only affect columns that are not included in the synchronization.

Say you have a table: `User` with the following columns
| `Id` |  `Firstname` | `Lastname` | `Password` |

Let's assume we exclude the column `Password` from sync.

When the user later changes the password, by default, 
- this change would still update the tracking table and the user will be re-synchronized to all devices although no relevant data has changed.
- Moreover, this could cause unnecessary conflicts and data loss: For example, client A changes the `Firstname` of a user while the `Password` is changed on the server.

During the next synchronization, this would be handled as a sync conflict with `ServerWins`, so the new `Firstname` would get lost.

So we changed our trigger on the serverside (MSSQL) like this:


```SQL 
ALTER TRIGGER [dbo].[Users_update_trigger] ON [dbo].[Users] FOR UPDATE AS

-- we need to make sure that an update only counts, if an actuall synchronized column changes

if exists (select i.id from inserted as i inner join deleted as d on i.Id = d.id
-- only trigger this if a tracked column of the table has changed
-- => IGNORE PASSWORD
where i.Id <> d.Id OR i.Firstname <> d.Firstname OR i.Lastname <> d.Lastname)
begin

UPDATE [side] 
SET 	[update_scope_id] = NULL -- since the update if from local, it's a NULL
	,[update_timestamp] = @@DBTS+1
	,[last_change_datetime] = GetUtcDate()

FROM [Users_tracking] [side]
JOIN INSERTED AS [i] ON [side].[Id] = [i].[Id]

end

```

Do you agree?

If not, we will still provide our own trigger using the Interceptor approach.
However, it would be awesome if that came out-of-the-box.

Mimetis:
Make totally sense !
I will update the trigger as soon as possible 👍 

But... (sorry there is always a *but* ;) )

To illustrate, I'm working on a sample table `Product` where filtered columns are `Name`, `Color` and `Size` :

Regarding your `Product_Update_trigger` you will encounter a weird behavior if you have multiple rows when your trigger is executed.

To illustrate, here is a quick demo (without any modification on the trigger, just to emphase the multiple rows behavior)

`Product_update trigger`: I used the `@@rowcount` property to see how many rows are present in the `inserted` temp table:

``` sql
ALTER TRIGGER [dbo].[Product_update_trigger] ON [dbo].[Product] FOR UPDATE AS

print 'rows count :' + convert(varchar, @@ROWCOUNT)

UPDATE [side] 
SET 	[update_scope_id] = NULL -- since the update if from local, it's a NULL
	,[last_change_datetime] = GetUtcDate()

FROM [tProduct] [side]
JOIN INSERTED AS [i] ON [side].[ProductID] = [i].[ProductID]

```

Quick demo:

``` sql 
Insert into Product (Name) 
Values ('Product 1') -- Will print Number or rows 1

Insert into Product (Name)
Select 'Product 1' Union Select 'Product 2' -- Will print Number or rows 2

```
So here is my version of the `Product_update_trigger` to handle this scenario, using a modified version of your idea:

``` sql
ALTER TRIGGER [dbo].[Product_update_trigger] ON [dbo].[Product] FOR UPDATE AS

print 'rows count :' + convert(varchar, @@ROWCOUNT)

UPDATE [side] 
SET 	[update_scope_id] = NULL -- since the update if from local, it's a NULL
	,[last_change_datetime] = GetUtcDate()

FROM [tProduct] [side]
JOIN INSERTED AS [i] ON [side].[ProductID] = [i].[ProductID]
LEFT JOIN DELETED [d] ON [i].[ProductID] = [d].[ProductID]
WHERE ([i].[Name] <> [d].[Name] or [i].[Color] <> [d].[Color] or [i].[Size] <> [d].[Size]);

```
For information, here is the `Product_insert_trigger`:

``` sql
ALTER TRIGGER [dbo].[Product_insert_trigger] ON [dbo].[Product] FOR INSERT AS

print 'Number of rows ' + convert(varchar, @@rowcount)


-- If row was deleted before, it already exists, so just make an update
UPDATE [side] 
SET  [sync_row_is_tombstone] = 0
	,[update_scope_id] = NULL -- scope id is always NULL when update is made locally
	,[last_change_datetime] = GetUtcDate()
FROM [tProduct] [side]
JOIN INSERTED AS [i] ON [side].[ProductID] = [i].[ProductID]
LEFT JOIN DELETED [d] ON [i].[ProductID] = [d].[ProductID]
WHERE ([i].[Name] <> [d].[Name] or [i].[Color] <> [d].[Color] or [i].[Size] <> [d].[Size]);

INSERT INTO [tProduct] (
	 [ProductID]
	,[update_scope_id]
	,[sync_row_is_tombstone]
	,[last_change_datetime]
) 
SELECT
	 [i].[ProductID]
	,NULL
	,0
	,GetUtcDate()

FROM INSERTED [i]
LEFT JOIN [tProduct] [side] ON [i].[ProductID] = [side].[ProductID]
LEFT JOIN DELETED [d] ON [i].[ProductID] = [d].[ProductID]
WHERE [side].[ProductID] IS NULL
AND ([i].[Name] <> [d].[Name] or [i].[Color] <> [d].[Color] or [i].[Size] <> [d].[Size])

```

What do you think ?

gentledepp:
> Regarding your Product_Update_trigger you will encounter a weird behavior if you have multiple rows when your trigger is executed.

UUH! I see!
So my bug is the following:
In case I update multiple `User` rows, and 
- row 1 only updates the `Password`, so it should be ignored
- row 2 updates the `Firstname`, so it should alter the timestamp in the `Users_tracking` table,
**my** `Users_update_trigger` would update the timestamp for both - row 2 **and** row 1 - which is total nonsense.

Your approach is much better! 🎉

### BUT ;-)

I assume you do not have to include the filter in the `_insert_trigger`s.

- When an insert happens, there is always one column, that changes **and** is relevant: The primary key ;-)
The same applies for `_delete_triggers` of course.

- Even more important: When an insert happens, there is never any row in the `DELETED` temp table. (At least not to my knowledge) So while your left outer join will not make any semantical difference, it still does unnecessary work.

gentledepp:
One short note for 
## 1 HttpClient
the following article explains why instances of `HttpClient` should be shared and potentially be a singeton for the whole app lifetime
https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests

Mimetis:
## Filter columns

Just encountered a weird bug for the `Update_trigger`

Here is the new version :
``` sql
ALTER TRIGGER [dbo].[Product_update_trigger] ON [dbo].[Product] FOR UPDATE AS

print 'rows count :' + convert(varchar, @@ROWCOUNT)

UPDATE [side] 
SET 	[update_scope_id] = NULL -- since the update if from local, it's a NULL
	,[last_change_datetime] = GetUtcDate()

FROM [tProduct] [side]
JOIN INSERTED AS [i] ON [side].[ProductID] = [i].[ProductID]
LEFT JOIN DELETED [d] ON [i].[ProductID] = [d].[ProductID]
WHERE ([i].[Name] <> [d].[Name] or [i].[Color] <> [d].[Color] or [i].[Size] <> [d].[Size] or
[i].[ProductCategoryId] <> [d].[ProductCategoryId]);
```

The problem is **When** you update a value to **Null** the operator **<>** is not working...

For instance:

``` sql
Update SalesLT.Product
Set ProductCategoryID = 'AAABBB'
where ProductID = 'A1B39DA0-D03F-419D-A052-1A2400018D60'

(1 row affected) -- Trigger fired, tracking table updated
(1 row affected) -- Product updated

Update SalesLT.Product
Set ProductCategoryID = NULL
where ProductID = 'A1B39DA0-D03F-419D-A052-1A2400018D60'

(0 row affected) -- Trigger NOT fired, tracking table NOT updated
(1 row affected) -- Product updated
```

Here a new version, using [NULLIF](https://docs.microsoft.com/fr-fr/sql/t-sql/language-elements/nullif-transact-sql)  and [ISNULL](https://docs.microsoft.com/fr-fr/sql/t-sql/functions/isnull-transact-sql) :

``` sql
ALTER TRIGGER [dbo].[Product_update_trigger] ON [dbo].[Product] FOR UPDATE AS

print 'rows count :' + convert(varchar, @@ROWCOUNT)

UPDATE [side] 
SET 	[update_scope_id] = NULL -- since the update if from local, it's a NULL
	,[last_change_datetime] = GetUtcDate()

FROM [tProduct] [side]
JOIN INSERTED AS [i] ON [side].[ProductID] = [i].[ProductID]
LEFT JOIN DELETED [d] ON [i].[ProductID] = [d].[ProductID]
WHERE (
  ISNULL(NULLIF([d].[Name], [i].[Name]), NULLIF([i].[Name], [d].[Name])) IS NOT NULL
OR ISNULL(NULLIF([d].[Size], [i].[Size]), NULLIF([i].[Size], [d].[Size])) IS NOT NULL
OR ISNULL(NULLIF([d].[Color], [i].[Color]), NULLIF([i].[Color], [d].[Color])) IS NOT NULL
OR ISNULL(NULLIF([d].[ProductCategoryID], [i].[ProductCategoryID]), NULLIF([i].[ProductCategoryID], [d].[ProductCategoryID])) IS NOT NULL)
```
Now, I need to find the same thing for **mysql** and **sqlite** :o

gentledepp:
## HttpClient
> With this new version, you can inject your own instance, is it not enough ?

Sure it is. I just wanted to share some background information with you ;-)

mdonogma:
Ability to sync read only one way with a sql view.

That way I’m not always constrained to physical table on server but client still has read only views of specific data needed by app.

Without this I have to write additional code to get the sql view data to the client.

Mimetis:
Sorry @mdonogma, today sync is working only on tables
One of the main reason is because the fx uses heavely some table statements.

mdonogma:
All good, love your work! I wouldn’t waste to much time on backward compatibility with 0.3 reason. Developers can simply plan upgrades into their releases. Example: app has cutoff date for all changes to be synchronised with server. Users are then forced to upgrade via in app version check which ends switches them to new backend sync URL which links to 0.4 changes.



It is more important for 0.4 to be a very stable release that can scale and so far I can see there are a ton of changes developers can benefit from. I’m going to give it a good scale test with 30 million records on server table and filter to ensure client only syncs with approx 10 000 records in total. If a scenario like that works on 0.4 you have a winner.

Mimetis:
Hey guys;

I've just add a new **Sql** provider that uses the **Change Tracking** feature, from SQL Server.
This provider is called `SqlSyncChangeTrackingProvider`.

What does it mean ?
- No more tracking tables in your database
- No more triggers on your tables in your database
- Metadatas retention managed by SQL Server itself 
- Some restrictions on filters (especially on Delete statement with a filter)

To be able to use `SqlSyncChangeTrackingProvider` on your database, do not forget to active the **Change Tracking** on your database :

``` SQL
ALTER DATABASE AdventureWorks  
SET CHANGE_TRACKING = ON  
(CHANGE_RETENTION = 14 DAYS, AUTO_CLEANUP = ON)  
```

You don't have to activate **Change Tracking** on each table. It will be enabled by the `Dotmim.Sync` framework.

It's an early preview, but I really want to release the `v0.4` with this new provider !

Please test it, and give me some feedbacks :)

PR #181, opened 2020-01-09
MySql trigger builder fix

lordofffarm:
Hi,

I was just trying out the current filters implementation on MySql and encountered this bug where triggers are not properly generated.

For example:
**Insert trigger before fix:**
insert1

**Insert trigger after fix:**
insert2

Same thing for update and delete triggers.


Mimetis:
Hey @lordofffarm !
Thanks for this fix !
I have to admit, I haven't done enough tests on the `MySql` provider...

I merge this PR with the master branch, and I have already integrated them in the next version `v0.4`.

If you're testing the `Dotmim.Sync` framework, you should consider to test the next version `v0.4` that you can see on the branch #177

Issue #222, opened 2020-02-20
Feature: Use hashes to ensure integrity of uploaded/downloaded batch files


gentledepp:
Hi!

We recently had an issue in our system, that a blob value of one of our synced rows (we use a blob column for some stuff) was "corrupted".

One of the reasons for this could be that there was a small glitch during the transmittion of batch files that caused some bits in that blob to be swapped - resulting in that error.
When sending data over unstable networks, you face those issues from time to time.

What would be great is if 
- Dotmim.Sync actually calculated the hash of the batch file it tries to transmit 
- and adds that hash as a header to the request/response
- the receiving end then re-calculates the hash
- and if it does not match the header, it discards the data and errors.

Hashing big data sets is not that performance intensive as you might think.
We have a simple .netstandard2 compatible Hasher that can output a hash from a stream as Base64 string (suitable for HTTP headers ;-))

What do you think?
I think it would be a great addition to 0.4!

```csharp
using System;
using System.IO;

 public static class Hashalgorithm
    {
        public static class SHA256
        {
            public static byte[] Create(byte[] data)
            {
                using (var stream = new MemoryStream(data))
                {
                    return Create(stream);
                }
            }

            public static byte[] Create(Stream stream)
            {
                using (var sha256 = System.Security.Cryptography.SHA256.Create())
                {
                    return sha256.ComputeHash(stream);
                }
            }

            public static string CreateAsBase64String(Stream stream)
            {
                return Convert.ToBase64String(Create(stream));
            }
        }
    }

```

To remain backward compatible, the receiving end can check this header only if it is available in the request/response.
If not, it just skips the hash-check ;-)

Mimetis:
hey, that seems legit !
I'm working today on `Dotmim.Sync` so I will take a look on it now :)
I'll let you know when it's tested and implemented

gentledepp:
Ok great. But no hurry needed - we are still on 0.3 ;-)

PR #223, opened 2020-02-20
Add a hash verification 

Mimetis:
See #222 for more info on this feature.
cc @gentledepp


gentledepp:
Hi!
Unfortunately, I struggle to see the complete contents of a file you changed. On DevOps that works like a charm. Github seems to not support this.
Therefore, I do not get the whole picture (as I only see fragments of a file), but it is looking good! :-)

PR #260, opened 2020-04-21
Modify Nullable parameter syntax.

HamzaRabah:
Hi @Mimetis 
We have started to use and implement this great library with SQL Server 2008R2 (with change tracking)
Unfurtanitly, creating the procedure for the update table doesn't work with tables that contain nullable columns because of a syntax error.
Please have a look,
Thanks.


Mimetis:
Hey @HamzaRabah !
Thanks for the PR.
Unfortunatelly, it's not working and almost all the tests are failing.

See the [logs on Azure Devops here](https://dev.azure.com/dotmim/Dotmim.Sync/_build/results?buildId=382&view=logs&j=5b51f4fd-f10c-5ba1-4461-7694e6f5956e&t=4d591615-0090-5a07-2128-18a215b015cf)

I've made a test, and for example, here is the script used for generating the `scope_info` table, that fail, obviously:

``` sql
CREATE PROCEDURE [dbo].[Customer_changes]
	@sync_min_timestamp bigint = 0,
	@sync_scope_id uniqueidentifier = NULL = NULL
AS
BEGIN
SELECT DISTINCT	[side].[CustomerID], 
	[base].[FirstName], 
	[base].[LastName], 
	[side].[sync_row_is_tombstone], 
	[side].[update_scope_id] 
FROM [Customer] [base]
RIGHT JOIN [Customer_tracking] [side] ON [base].[CustomerID] = [side].[CustomerID]
WHERE (
	[side].[timestamp] > @sync_min_timestamp
	AND ([side].[update_scope_id] <> @sync_scope_id OR [side].[update_scope_id] IS NULL)
)
``` 
As you can see, this line is making the creation failing:

``` sql
	@sync_scope_id uniqueidentifier = NULL = NULL
```

Can you provide more context. What is the situation for you, where **DMS** is failing ?

HamzaRabah:
Hey @Mimetis
This is the error message:
`Incorrect syntax near the keyword 'NULL'.
Must declare the scalar variable "@sync_scope_id".`
and I've attached the [StackTrace](https://github.com/Mimetis/Dotmim.Sync/files/4515762/stacktrace.txt),
This issue happened when syncing tables that's contains nullable columns only in SQL Server 2008R2, Because the generated SQL script for creating the procedure add `NULL` for nullable columns parameters instead of `= null`, So the first syntax is not valid in SQL server 2008R2 but it's working fine on newer versions.

Mimetis:
Thanks for the feedback

I've created a virtual machine with SQL Server 2008R2 and I'm able to reproduce the bug.
You're right !

I think a solution could be :

``` cs
if (param.Value != null)
 stringBuilder3.Append($" = {param.Value}");
else if (param.IsNullable)
 stringBuilder3.Append(" = NULL");

```
I'm actually trying this solution and if it's working as expected, I will push a new version asap. (probably later today)

Mimetis:
Can you try with the last version `v0.5.3` ? 
It's already available on **nuget**: https://www.nuget.org/packages?q=dotmim.sync

HamzaRabah:
Hi @Mimetis 
Thanks a lot for your efforts and fast respond,
It's working perfectly fine,
Thanks again for your work and this great library.

Mimetis:
Thanks to you as well @HamzaRabah to make this library more compatible ;)

PR #266, opened 2020-04-30
First server reprovision fails due to non existing entry in serverScopeInfo table

jeremy-waguet:


Mimetis:
I would have made it a little bit differently.

In you case, you're affecting a new setup / schema, but the migration engine will run anyway, and we don't want it, because it will be the responsability of your `ProvisionAsync` call.

In my opinion, a better solution would have been to just return the serverscope "as is" if it's a new one:

``` csharp
// if serverscopeinfo is a new, because we never run any sync before, just return serverscope empty
if (serverScopeInfo.Setup == null && serverScopeInfo.Schema == null)
    return serverScopeInfo;

```
In context:

``` csharp
// Get scope if exists
(ctx, serverScopeInfo) = await this.Provider.GetServerScopeAsync(ctx, this.Options.ScopeInfoTableName, this.ScopeName, connection, transaction, cancellationToken, progress).ConfigureAwait(false);

// if serverscopeinfo is a new, because we never run any sync before, just return serverscope empty
if (serverScopeInfo.Setup == null && serverScopeInfo.Schema == null)
    return serverScopeInfo;

// Compare serverscope setup with current
if (serverScopeInfo.Setup != this.Setup)
{
    this.logger.LogDebug(SyncEventsId.GetServerScope, $"[{ctx.SyncStage}] database {connection.Database}. serverScopeInfo.Setup != this.Setup. Need migrations ");

    SyncSet schema;
    // 1) Get Schema from remote provider
    (ctx, schema) = await this.Provider.GetSchemaAsync(ctx, this.Setup, connection, transaction, cancellationToken, progress).ConfigureAwait(false);
    schema.EnsureSchema();

   ....
   ....
```
Once implemented like that, we won't have any error raised if we try to call the `ProvisionAsync()` method when nothing have been already created on the server or client database.

The calling is something like that (note the comments to understand):

``` csharp
// This method is useful if you want to provision by yourself the server database
// You will need to :
// - Create a remote orchestrator with the correct setup to create
// - Get the server scope that will contains serialized version of your scope / schema
// - Provision everything
// - Save the server scope information

// Create a server orchestrator used to Deprovision and Provision only table Address
var remoteOrchestrator = new RemoteOrchestrator(serverProvider, options, setup);

// Now we need to save this new schema to the serverscope table
// get the server scope again
var serverScope = await remoteOrchestrator.GetServerScopeAsync();

// Server scope is created on the server side.
// but Setup and Schema are both null, since nothing have been created so far
//
// serverScope.Setup = null;
// serverScope.Schema = null;
//
// Provision everything needed (sp, triggers, tracking tables)
// Internally provision will fectch the schema a will return it to the caller. 
var newSchema = await remoteOrchestrator.ProvisionAsync(SyncProvision.StoredProcedures | SyncProvision.Triggers | SyncProvision.TrackingTable);

// affect good values
serverScope.Setup = setup;
serverScope.Schema = newSchema;

// save the server scope
await remoteOrchestrator.WriteServerScopeAsync(serverScope);

```

your thought ?

jeremy-waguet:
Ok I agree with you. We should not provision internally. I'll do some test to see if it works well before I confirm it to you and update my PR in consequence.

Mimetis:
I've updated the master branch and I've added the documentation as well.
Check it out here : https://dotmimsync.readthedocs.io/Provision.html

You can test the last pre-release as well : https://www.nuget.org/packages/Dotmim.Sync.SqlServer/0.5.5-beta-0401

Let me know the results of your tests :)

jeremy-waguet:
@Mimetis Okay, it is working perfectly both on first run (initial EF migration) and with columns added !. So I am closing this PR and the related issue. Thank you !

PR #293, opened 2020-06-24
Mysql binary fix

lordofffarm:
In MySql when column type is "binary", Dotmim.Sync throws an exception.
Turns out this type is not handled properly.

### DIFF
--- a/Projects/Dotmim.Sync.MySql/Builders/MySqlDbMetadata.cs
+++ b/Projects/Dotmim.Sync.MySql/Builders/MySqlDbMetadata.cs
@@ -55,7 +55,9 @@ namespace Dotmim.Sync.MySql.Builders
                     case "tinytext":
                         return 0;
                     case "varbinary":
+                        return 65535;
                     case "binary":
+                        return 255;
                     case "text":
                     case "nchar":
                     case "nvarchar":
@@ -85,7 +87,7 @@ namespace Dotmim.Sync.MySql.Builders
                 case DbType.AnsiStringFixedLength:
                     return MySqlDbType.VarChar;
                 case DbType.Binary:
-                    return MySqlDbType.Binary;
+                    return MySqlDbType.VarBinary;
                 case DbType.Boolean:
                     return MySqlDbType.Bit;
                 case DbType.Byte:
@@ -232,8 +234,8 @@ namespace Dotmim.Sync.MySql.Builders
                     case "json":
                     case "tinytext":
                         return string.Empty;
-                    case "varbinary":
-                    case "binary":
+                    case "varbinary" when maxLength <= 65535:
+                    case "binary" when maxLength <= 255:
                     case "text":
                     case "nchar":
                     case "nvarchar":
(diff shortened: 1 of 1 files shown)


Mimetis:
Hey @lordofffarm 

Seems we have issues with the tests:
https://dev.azure.com/dotmim/Dotmim.Sync/_build/results?buildId=431&view=logs&j=f28b2766-c77f-5700-fbec-7a15fb52f368&t=885d2a0d-2ec0-5a54-b95b-662ea45a4f39

lordofffarm:
Hi @Mimetis .
I am not sure if the changes are right, but the tests are running successfully now...

Projects/Dotmim.Sync.MySql/Builders/MySqlDbMetadata.cs
```diff
@@ -43,22 +45,26 @@ public override int GetMaxLengthFromOwnerDbType(object dbType, int maxLength)
...
                         return 0;
-                    case "TEXT":
-                    case "NCHAR":
-                    case "NVARCHAR":
-                    case "VARCHAR":
-                    case "CHAR":
-                    case "ENUM":
-                    case "SET":
+                    case "varbinary":
+                        return 65535;
```
Mimetis, line 58: I would have done a little different here, to handle the column maxlength:
(macro code)
if maxLength > 0 return Math.Min(maxLength, 65535) else return 65535;

Projects/Dotmim.Sync.MySql/Builders/MySqlDbMetadata.cs
```diff
@@ -43,22 +45,26 @@ public override int GetMaxLengthFromOwnerDbType(object dbType, int maxLength)
...
+                    case "binary":
+                        return 255;
```
Mimetis, line 60: same comment as varbinary

Projects/Dotmim.Sync.MySql/Builders/MySqlDbMetadata.cs
```diff
@@ -439,6 +447,8 @@ public override bool IsTextType(string typeName)
             string lowerType = typeName.ToLowerInvariant();
             switch (lowerType)
             {
+                case "varbinary":
+                case "binary":
```
Mimetis, line 451: varbinary / binary are considered as text in mysql ?

lordofffarm: The documentation says

> The BINARY and VARBINARY types are similar to CHAR and VARCHAR, except that they store binary strings rather than nonbinary strings. That is, they store byte strings rather than character strings. This means they have the binary character set and collation, and comparison and sorting are based on the numeric values of the bytes in the values.

They store binary strings, and comparison and sorting are based on the numeric values of the bytes in the values

[Docs](https://dev.mysql.com/doc/refman/8.0/en/binary-varbinary.html)

lordofffarm:
Also, this code with the "8000" looks like it came from SqlDbMetadata?

```
public override int ValidateMaxLength(string typeName, bool isUnsigned, bool isUnicode, long maxLength)
{
      Int32 iMaxLength = maxLength > 8000 ? 8000 : Convert.ToInt32(maxLength);
      return iMaxLength;
}
```

Should this be different for MySql?

Projects/Dotmim.Sync.MySql/Builders/MySqlDbMetadata.cs
Mimetis: Ok, Sounds good to me !

Mimetis:
Thx !

PR #315, opened 2020-08-26
Added support for "REAL" data type in SQLite

vunder:
Current implmenetation missing one of the basic SQLite data type - `REAL` ([SQLite data type](https://www.sqlite.org/datatype3.html))
Added convertion `SQLite REAL` <-> `C# Double`


vunder:
Any comments?
REAL is one of the basic SQLite Data Types. Considering database compatibility between several platforms, for `System.Double` C# type better use SQLite `REAL`, rather than `NUMERIC`

Mimetis:
hey @vunder 
Thanks for the PR, I will commit the PR as soon as possible
I just need a little bit of time to review it, but as far as I see, it's OK and tests are OK too.
SO almost GOOD TO GO 👍 

Thanks for your contribution, really appreciate !

PR #331, opened 2020-09-07
Sqlite performance

Mimetis:


vunder:
Just some performance comments, regarding changes.
Client (UWP) and WebApi server are running locally.
Client has ~35K records within 20+ tables.
Sync in `Reinitialize` mode currently takes ~55 seconds. Before this PR it took almost 7 minutes.

Issue #300, opened 2020-07-22
0.4.3 Sqlite file gets out of sync


dgrothman:
I am running dotmim 0.4.3 currently and has been running well, I have ran into an issue with a single instance where the sqlite database on the client has information that is not on the server, but the sync doesn't "see" it and therefore it never gets pushed to the cloud database (azure sql). If I try and sync the agent using syncType.ReinitializeWithUpload the data gets destroyed. Is there any way I can force it to recheck the table the data is in and upload it? I have the sqlite file backed up, so I can try making changes in the sqlite file itself if needed, but I don't understand how the tracking tables work.

gentledepp:
TLNR;
You need to do a normal sync in those cases, but reset the local timestamp to 0 (lastSyncTimestamp in _scopeInfo table), so that all changes are resent. This will cause a whole lot of sync conflicts and take a while, but it will work.


If you are using "ReinitializeWithUpload", the client will only send *pending* changes to the server. Thereafter it will delete all local data (thus destroying your forgotten local changes) and re-download all data from the server.
Due to a misconception in the sync framework, the client can "overlook" pending changes and never send them.

Have a look at my explanation titled 
#### 4 Warning using SqliteProvider due to transaction-overlappings (no timestamp exists on sqlite)
here: https://github.com/Mimetis/Dotmim.Sync/issues/255


# Our current solution
We use a multiple reader/single writer lock in C# to ensure, that two sqlite transactions cannot overlap. The (single) writer access is taken by the dotmim.sync while all other transactions (any application access to the database) uses the reader lock.
This ensures that dodmim.sync cannot collect any local changes, while some other transaction has not yet committed.

# The best solution
...would be if dotmim.sync would opt to switch back to the same change detection approach that the original Microsoft Sync Framework used locally: (As far as I remember) it had a `dirty` column in the `_tracking` tables and did not use a `lastSyncTimestamp`. That way, even if transactions overlapped, a subsequent sync would detect `dirty` rows and send them.
@Mimetis  what do you think? This causes real headache. As SQLite does not have a real `timestamp` and no `min_active_rowversion` I guess this is the only real option. 
I guess it won't hurt, if we'd introduce an additional `dirty` column **only for SQLite**.

Mimetis:
@dgrothman 

A row that "*is tracked*" has its own corresponding row in the `tracking_table`.

Let's say for example, in the **Customer** table, for one row:
[...]

PR #353, opened 2020-10-13
#300 - fixing issue causing dataloss when server session is lost whil…

gentledepp:
…e client sends additional batches

# 1 - solves data loss issue 

The issue is, that the `WebServerOrchestrator` just creates a new `clientBatchInfo` if it is null.
The issue with this is: If you send changes in batches and the server session is lost (e.g. due to an ApplicationPool recycle or because the Redis cache was cleared), it will just forget all the batches that were already sent during the ongoing session!

See the following sequencediagram for clarification:
```mermaid
sequenceDiagram
    participant Client
    participant Server
    loop GetLocalChanges
        Client->>Client: create 2 batch files (b0, b1, b2)
    end
    Client->>Server: Send b0
    Note right of Server: session is lost <br/>=> clientBatchInfo is null
    Client->>Server: Send b1
    Client->>Server: Send b2
    loop ApplyChanges
         Server->>Server: deserialize all batches and apply to database
         Note right of Server: Only inserts b1 and b2!<br/>b0 changes are forever lost!
    end

``` 

The only thing I had to add to fix this is a null-check:
```diff
            // ------------------------------------------------------------
            // FIRST STEP : receive client changes
            // ------------------------------------------------------------

+            // ensure that the blientBatchInfo is still available in the session - it **must** only be null when the batchIndex is 0!!
+            if (sessionCache.ClientBatchInfo == null && httpMessage.BatchIndex != 0)
+                throw new SyncException("Session loss: No batchPartInfo could found for the current sessionId. It seems the session was lost. Please try again.");

            // We are receiving changes from client
            // BatchInfo containing all BatchPartInfo objects
            // Retrieve batchinfo instance if exists
            // Get batch info from session cache if exists, otherwise create it
            if (sessionCache.ClientBatchInfo == null)
                sessionCache.ClientBatchInfo = new BatchInfo(clientWorkInMemory, Schema, this.Options.BatchDirectory);

```

Interestingly, your code **does** check whether the session is empty when the client is **requesting** additional batches.
So I just added a test to cover the fact that this will also not break in the future.

# 2 `HttpMessageSendChangesRequestArgs` takes a `byte[]` - what for?
I had to implement "OnReceiveChanges" interception
Why is the parameter `HttpMessageSendChangesRequestArgs` of `OnSendChanges` taking the binary data as a parameter?

I would argue, that when I do intercept this call, I would like to see what changes are sent.
But in the current implementation, I only do get a byte array.
The only thing I could think of was to manipulate this array to e.g. add another layer of compression/encryption. Is that the intention?
As I did not see any real sense in that, I removed this.
Instead (and also in order to create an integration test for this scenario) I changed the constructor to take a `HttpMessageSendChangesRequest` as a parameter instead of the `byte[]`:
```
public class HttpMessageSendChangesRequestArgs : ProgressArgs
    {
        public HttpMessageSendChangesRequestArgs(HttpMessageSendChangesRequest request) 
		{
			...
		}
	}

```

Likewhise, I had to add `HttpMessageSendChangesRequest` to the `HttpMessageSendChangesRequestArgs`

Additionally, the Interception happens **before** the changes are serialized.
That way, the interceptor can actually *modify* the changes sent to the server!

# 3 - Added integration test to ensure deferred transaction are never used by SqliteSyncProvider
Additionally, I added a specific integration test that ensures that the SqliteSyncProvider does **not** use deferred transactions when selecting local changes as this can also lead to data loss in the future!

### DIFF
--- a/Projects/Dotmim.Sync.Web.Client/HttpMessageArgs.cs
+++ b/Projects/Dotmim.Sync.Web.Client/HttpMessageArgs.cs
@@ -1,13 +1,4 @@
-﻿using Dotmim.Sync.Batch;
-using Dotmim.Sync.Builders;
-using System;
-using System.Collections.Generic;
-using System.Text;
-using Dotmim.Sync.Enumerations;
-using System.Runtime.Serialization;
-using System.Data.Common;
-
-namespace Dotmim.Sync.Web.Client
+﻿namespace Dotmim.Sync.Web.Client
 {
     public class HttpMessageSendChangesResponseArgs : ProgressArgs
     {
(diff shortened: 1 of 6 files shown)


Mimetis:
THIS. IS. AWESOME !
Thanks @gentledepp  for this PR.
I'll make a quick review this week and will merge it asap !

Mimetis:
> I had to implement "OnReceiveChanges" interception
> Why is the parameter HttpMessageSendChangesRequestArgs of OnSendChanges taking the binary data as a parameter?
> The only thing I could think of was to manipulate this array to e.g. add another layer of compression/encryption. Is that the intention?

Indeed, but since we have `Converters ` and `Serializers`, I guess this interceptor is kind of obsolete now
Your implementations seems more accurate, thanks :)

PR #558, opened 2021-06-03
Minor fixes

vunder:
Fixes in RyncResults that may cause debug errors because NullReferenceException
AfterDeserializedRows was not called while HTTP-sync
Correct sync on server in case there is not body hash in request headers

### DIFF
--- a/Projects/Dotmim.Sync.Core/SyncResult.cs
+++ b/Projects/Dotmim.Sync.Core/SyncResult.cs
@@ -34,12 +34,12 @@ namespace Dotmim.Sync
         /// <summary>
         /// Gets the number of changes applied on the client
         /// </summary>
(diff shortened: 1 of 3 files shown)


vunder:
I see that few tests are falling because of error in `DateConverter.AfterDeserialized` that should convert ticks to DateTime. But in fact, values there are DateTime instead of Int64

Mimetis:
Thanks a ton @vunder for these fixes !!

Mimetis:
BTW, I found the bug on Converter :)

vunder:
@Mimetis Yes, that's why some tests are failing. And I wrote about it.
There are unit tests for data conversion. However, `DateConverter` in tests does not consider that

Mimetis:
I missed a serialization process in a specific case, that should fine now :)
Thanks again for the various fixes, it's helping a lot !!!

PR #570, opened 2021-07-07
Fix UnsupportedColumnNameException for the Column name Timestamp (SQLServer)

tuyndv:
Fix bug for : UnsupportedColumnNameException: The Column name Timestamp is not allowed. Please consider to change the column name.

### DIFF
--- a/Projects/Dotmim.Sync.SqlServer/Builders/SqlBuilderTable.cs
+++ b/Projects/Dotmim.Sync.SqlServer/Builders/SqlBuilderTable.cs
@@ -244,6 +244,10 @@ namespace Dotmim.Sync.SqlServer.Builders
                 var name = c["name"].ToString();
                 var maxLengthLong = Convert.ToInt64(c["max_length"]);
 
+                if (name.ToLower().Equals("timestamp"))
+                {
+                    name = $"[{name}]";
+                }
                 //// Gets the datastore owner dbType 
                 //var datastoreDbType = (SqlDbType)sqlDbMetadata.ValidateOwnerDbType(typeName, false, false, maxLengthLong);
 


Mimetis:
thanks @tuyndv for this PR

Looking at your code, you are trying to break the rule on timestamp column name by replacing it with [timestamp]
In my opinion, it's not the best idea here :)

The validation of column names are here : https://github.com/Mimetis/Dotmim.Sync/blob/437c4da494a6e6d6d213217a37f1bfabe502ea60/Projects/Dotmim.Sync.Core/Orchestrators/BaseOrchestrator.Schema.cs#L189-L204

This rule enforces the naming, because we don't want to have such columns in a `select` that will conflict with the columns from the tracking tables

Honestly, I think we can have a better way to enforce the consitensy between tracking tables columns and user tables columns, if we don't have conflicts on primary keys (and not on others columns)

I will work on this.
If I'm right, you will be able to use `timestamp` as a column name (IF it's not your primary key)

I will not merge this PR, if my solution is working, otherwise we will continue the discussion here.

Thanks a ton !

tuyndv:
Thanks @Mimetis

Mimetis:
In the last commit, I tried to be more specific about column names.

for regular columns (not primary keys) here are the forbidden column names:

``` cs
columnNameLower == "sync_scope_id"
|| columnNameLower == "changeTable"
|| columnNameLower == "sync_min_timestamp"
|| columnNameLower == "sync_row_count"
|| columnNameLower == "sync_force_write"
|| columnNameLower == "side_update_scope_id"
|| columnNameLower == "side_timestamp"
|| columnNameLower == "side_sync_row_is_tombstone"
```

for primary column name, here are the forbidden ones (including forbidden names from regular columns):

``` cs
columnNameLower == "sync_scope_id"
|| columnNameLower == "sync_scope_name"
|| columnNameLower == "update_scope_id"
|| columnNameLower == "timestamp"
|| columnNameLower == "timestamp_bigint"
|| columnNameLower == "sync_row_is_tombstone"
|| columnNameLower == "last_change_datetime"
```

As you can see, I changed `timestamp` to `side_timestamp`.
`timestamp` is really used often, so that makes sense to have a different names in stored procedures.

That being said, the new convention will be part of version `v0.8.1` that is not yet published but already on master branch.
You can test it if you want, by cloning the code.

DO NOT try it on a database where you have already a sync in place.
Try on a new fresh database, to allow DMS to create the good stored procedures

tuyndv:
Dear Mimetis,

It is working for me. Thanks @Mimetis

PR #597, opened 2021-08-25
Patch for HttpRequestHandler

tdoerfler:
Some minor fixes for HttpRequestHandler I came across while testing. The first will fix an ObjectDisposedException when the client will retry to send a request again, cause the ByteArrayContent will be disposed on first failed request. Further I removed the call to responseStream.FlushAsync to avoid an InvalidOperationException as the stream is a read-only stream that doesn't need to be flushed.


Mimetis:
Thanks @tdoerfler for this quick fix !

Merged ;)

PR #660, opened 2021-12-14
Fix issue in sql change tracking when columns are updated that are not in the setup

kevindost:
The behaviour should be that when the sync setup contains columns, only those columns should be involved in the sync.

### DIFF
--- a/Projects/Dotmim.Sync.Core/Setup/SyncSetup.cs
+++ b/Projects/Dotmim.Sync.Core/Setup/SyncSetup.cs
@@ -89,6 +89,10 @@ namespace Dotmim.Sync
         /// </summary>
         public bool HasColumns => this.Tables?.SelectMany(t => t.Columns).Count() > 0;  // using SelectMany to get columns and not Collection<Column>
 
+        /// <summary>
+        /// Check if Setup has a table that has columns
+        /// </summary>
+        public bool HasTableWithColumns(string tableName) => this.Tables[tableName]?.Columns?.Count > 0;
 
         /// <summary>
         /// Check if two setups have the same local options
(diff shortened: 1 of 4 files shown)


kevindost:
The reason you see the following subquery in the select 

```
...
(CASE WHEN [SYS_CHANGE_OPERATION] = 'I' THEN (
SELECT [N].[SYS_CHANGE_COLUMNS]
FROM CHANGETABLE(CHANGES [Address], @next_sync_min_timestamp) AS [N]
WHERE [N].[AddressID] = [CT].[AddressID]
) ELSE [SYS_CHANGE_COLUMNS] END) AS [sync_change_columns]
FROM CHANGETABLE(CHANGES [Address], @sync_min_timestamp) AS [CT]
...
```
is to get changed columns for the next change version when the current version indicates it was an insert. This way we can compute if the record is eligable for synchronization since the last synchronization. Without this safeguard the following can happen:
1. The change version is 0, retrieved by the sync.
2. The sync inserts a record, the change version is updated to 1.
3. The sync finishes and stores change version 0.
4. The record that was previously inserted will have a column updated that is not part of the setup, change version is now 2.
5. A new sync starts, change version 0 is retrieved from the datastore and will be used to get all changes since that version 
6. Since that record's change version is 2 and since it's inserted since 0, all columns will count as changed and will be part of sync

kevindost:
@Mimetis pr should be final, if you have time can you review my latest changes?

Mimetis:
@kevindost Great.
I see you fixed the issue with the sync_row_is_tombstone in the stored procedures. good job.
Just waiting for the tests to be green, and we should be good to go in `v0.9.3`

kevindost:
@Mimetis Thanks, yes the `sync_row_is_tombstone` is a good thing to include for the custom filters.

Mimetis:
Sounds good to me, even if the generated sql seems to be really complicated :)

PR #676, opened 2022-01-05
Exception in SqlSyncAdapter.cs. Remove DateTimeOffset conversion and handle SmallDateTime better

### LINKED ISSUE #680: Exception in SqlSyncAdapter
Changing the value of DateTimeOffset when the rowValue < SqlDateMin throws exception when record.SetValue is called because the types are different

slagtejn:
Changing the value of DateTimeOffset when the rowValue < SqlDateMin throws exception when record.SetValue is called because the types are different.  There is no reason to be changing the data on DateTimeOffset since it can handle values lower than SqlDateMin.  Only DateTime & SmallDateTime have limits so I also adjusted the code to handle those.

Closes #680


Mimetis:
Thanks @slagtejn for this PR !
Sounds good to me

PR #722, opened 2022-03-31
Update WebClientOrchestrator.cs

### LINKED ISSUE #721: Upload more than 2 batches fails when using inside .NET 4.8
When using the library inside .net 4.8 there is a max of 2 connections per endpoint, but inside .net core it's not capped so this issue is not noticeable.  Since the SendAsync is using 'HttpCompletionOption.ResponseHeadersRead' we must either dispose the response or stream otherwise the connection will not close.  Uploading batches is done inside a foreach without disposing the response or reading the stream.  If more than 2 batches need to be uploaded it will hang forever.  The other option is to remove 'HttpCompletionOption.ResponseHeadersRead', but that was put in for memory and performance reasons.

https://github.com/dotnet/runtime/issues/30304

I made a fix and will submit a PR.

Fixed with PR #722

slagtejn:
Fixes #721


Mimetis:
Hello @slagtejn 
Thanks for this PR ! 

Since the next release is around the corner, I will close this PR, but your fix is already integrated here : https://github.com/Mimetis/Dotmim.Sync/commit/55a8b36f4e2c82076173b20f791c827147df1606

Thanks for your contribution :)

PR #768, opened 2022-08-01
Update session end to contain SyncResult

slagtejn:
Added SyncResult to the 'SessionEndArgs' because it's not possible to get the SyncResult in relation to the SyncContext like before.  The SyncContext is now being passed around everywhere and doesn't exist as GetContext() on the 'WebRemoteOrchestrator' anymore.  Adding this to the SessionEnd allows users to intercept the event and perform custom operations on the SyncResult in relation to the SyncContext such as a custom session ended event which inserts statistics into a custom tables based on the sync scope id which exists on the context.


Mimetis:
Thanks for this PR !
Indeed, it makes sense to have the SyncResult in the EndSession interceptor

PR #770, opened 2022-08-02
Fix localjsonserializer to handle datetimeoffsets

slagtejn:
Fix syncing DateTimeOffset columns.  A column that is type DateTimeOffset will be converted to local time and offset during de-serialization.  This means data is changing from the server -> client and the other way if data is being uploaded.  The data is being de-serialized as objects not concrete classes.  The default parsing of datetime is datetime and not datetimeoffset. This means all datetimeoffset values are being converted to local times.

This change will force dates to be de-serialized as datetimeoffsets and then convert datetime columns to datetime values.  The datetime values will have an unspecific kind so the data will not change.

If there is a better way to handle this then let me know.


Mimetis:
The DateTime trick you are using is really heavy on memory consumption, since you are making a copy of the array, plus you are doing some reflection to get the type

I think performances should be impacted here.

Before trying to fix it, I need to make some tests.
Dou you have any sample to share to reproduce the error, that I can use to investigate ?

slagtejn:
It's replacing the DateTimeOffset values in the array with DateTime values so it's not creating a new array.  Also the GetDataType() is not reflection, but a method on the column to determine the type based on the SQL column type which is an integer.
[CreateAdventureWorks_DateTimeOffset.zip](https://github.com/Mimetis/Dotmim.Sync/files/9252692/CreateAdventureWorks_DateTimeOffset.zip)
[CreateAdventureWorks_DateTimeOffset.zip](https://github.com/Mimetis/Dotmim.Sync/files/9252699/CreateAdventureWorks_DateTimeOffset.zip)

I attached a new CreateAdventureWorks database with the Address table with the ModifiedDate having a datetimeoffset.  Also the first record will insert a datetimeoffset value.  When syncing the data you will see the client will change the date and change the offset to your local timezone.

Mimetis:
The problem is coming from the **Json serializer options**
I've added a couple of options to handle _DateTimeOffset_, in [LocalJsonSerializer](https://github.com/Mimetis/Dotmim.Sync/blob/master/Projects/Dotmim.Sync.Core/Serialization/LocalJsonSerializer.cs)

Let me know if it's better now ? (using the last commit form master)

slagtejn:
Yes and it's also converting the classic DateTime to local on the client because it's being serialized as a DateTimeOffset with utc (offset 0).  This is why I had the code to convert the classic DateTime back to the correct type.

Mimetis:
I did not find a way to make it better than your solution.
So thanks a lot for this quick fix, It's merged !!!

Mimetis:
Oups we have a small conflict
Can you resolved it ?
Then I'll merge it

slagtejn:
Conflicts have been resolved

PR #773, opened 2022-08-04
Incorrect variable called in SynchronizeAsync

Davez69gto:
I'm going through your documentation and just noticed an incorrect variable name.  Nothing exciting.


Mimetis:
thanks ;)

PR #790, opened 2022-08-18
Sql command timeout

maprado-inetum:
Sql command timeout parametrized from SyncSetup for execute sql commands with a lot of data

### DIFF
--- a/Projects/Dotmim.Sync.Core/Orchestrators/BaseOrchestrator.ApplyChanges.cs
+++ b/Projects/Dotmim.Sync.Core/Orchestrators/BaseOrchestrator.ApplyChanges.cs
@@ -131,7 +131,7 @@ namespace Dotmim.Sync
                 return context;
 
             context.SyncStage = SyncStage.ChangesApplying;
-            
+
             var setupTable = scopeInfo.Setup.Tables[schemaTable.TableName, schemaTable.SchemaName];
 
             if (setupTable == null)
@@ -248,6 +248,8 @@ namespace Dotmim.Sync
                         // get the correct pointer to the command from the interceptor in case user change the whole instance
                         command = batchArgs.Command;
 
+                        command.CommandTimeout = scopeInfo.Setup.SqlCommandTimeout;
+
                         await this.InterceptAsync(new DbCommandArgs(context, command, connection, transaction), progress, cancellationToken).ConfigureAwait(false);
 
                         // execute the batch, through the provider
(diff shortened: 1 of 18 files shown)


Mimetis:
Hey @maprado-inetum 
Thanks for the PR

I understand that `CommandTimeout` is something that can be useful in some circumstances.
That being said, I think setting this property in the `SyncSetup` object is not the right place.

Let me explain:
- **`SyncSetup`** contains all the "_things_" that are, and must be, shared between the **Server** and all the **Clients**, like **tables** definition for example.
- **`SyncOptions`** contains all the "_things_" that are used by the local provider (**Server** or **Client**) and that should not be shared between all the sync participants, like `BatchDirectory` for example.
 
I think that `CommandTimeout` is something that should be placed as a `SyncOptions`'s property.
That way you can have a specific `CommandTimeout` for the server and let the default one for all clients.

Can you "move" the `CommandTimeout` property from `SyncSetup` to `SyncOptions`?
_In My Opinion_, it should be a nullable int property to let the provider default value i
Looking at the documentation for SqlServer or Sqlite , it seems that the default value is 30 (seconds)

And if you want to implement this option, it should be implemented in **all** `DbCommand` in `Dotmim.Sync.Core`.

Make sense?

maprado-inetum:
Hello and thanks for your quick review!!

I will continue to develop this feature using your suggestions ;)

maprado-inetum:
Changes done. I 'move' the SqlCommandTimeout parameter to SyncOptions object.
Regards

Mimetis:
Thanks a lot for the work !
That being said, I think you should consider some changes.

First of all, I think the naming of the property is not good.
You called it `Options.SqlCommandTimeout`, but this option is not only for **SQL**. It can be used with **MySql**, **MariaDB** or **Sqlite**.
I think you should rename it to `Options.CommandTimeout` or `Options.DbCommandTimeout`

Then I think this option should be a nullable int:

``` cs
/// <summary>
/// Gets or Sets the commands timeout in the sync
/// </summary>
public int? CommandTimeout { get; set; }
```

And default value should be Null. That way we are relying on the default value from the provider itself, instead of fixing an arbitrary value (of 30 sec in your case)

Finally, if we want to rely on default value of the provider, if we set a null value, we need to handle it on each command:

``` cs
if (Options.CommandTimeout.HasValue)
      command.CommandTimeout = Options.CommandTimeout.Value;
```

What do you think ?

maprado-inetum:
Like this? ;)

Mimetis:
Good to me :)

Little things to modify, but I will do it later

Thanks a lot for your contribution !

maprado-inetum:
Hi @Mimetis, ¿When do you deploy a new nuget release with this changes? Thanks!

Mimetis:
A big update is coming soon, and this PR will be part of it

maprado-inetum:
Can yo tell me the planning please?

Mimetis:
I'll do my best to push a new Release before end of this week (or maybe next week)

PR #803, opened 2022-08-31
Fix GetExistsScopeInfoTableCommand & GetExistsScopeInfoClientTableCommand to include schema

slagtejn:
Added a schema check since these tables could exist in a different schema.  An example would be the scope_info table is now the same name as the sync framework table.


Mimetis:
You can rename the `scope_info` table using the `SyncOptions.ScopeInfoTableName` property.

`DMS` is not using schema for `scope_info`, that's why there is no schema check.
How can this be useful ?

slagtejn:
ok I didn't realize this option existed.  Then yes this is not useful

PR #841, opened 2022-10-13
Fix #831

### LINKED ISSUE #831: TryConvertTo does not work on IOS when using dynamic for 'DateTime'
Hi @Mimetis

I've exact the same problem as mentioned in Issue #417 under Xamarin (iOS) but with '**DateTime**'.
After changing:

```csharp
else if (typeOfT == typeof(DateTime))
{
	if (DateTime.TryParse(value.ToString(), provider, DateTimeStyles.None, out DateTime dateTime))
		return (T)Convert.ChangeType(dateTime, typeOfT, provider);
	else if (typeOfU == typeof(long))
		return (T)Convert.ChangeType(new DateTime(value), typeOfT, provider);
	else
		return Convert.ToDateTime(value);
}
```
to:

**string valueStr = value.ToString();**
	if (DateTime.TryParse(**valueStr**, provider, DateTimeStyles.None, out DateTime dateTime))

```csharp
else if (typeOfT == typeof(DateTime))
{
	string valueStr = value.ToString();
	if (DateTime.TryParse(valueStr, provider, DateTimeStyles.None, out DateTime dateTime))
		return (T)Convert.ChangeType(dateTime, typeOfT, provider);
	else if (typeOfU == typeof(long))
		return (T)Convert.ChangeType(new DateTime(value), typeOfT, provider);
	else
		return Convert.ToDateTime(value);
}
```

It works perfect with the solution from @gaurakshay for me.
Could you please change that because I've never made an PR so far 😅😇

Thanks for your great work 😀!

symbiogenesis:
https://github.com/Mimetis/Dotmim.Sync/issues/831

### DIFF
--- a/Projects/Dotmim.Sync.Core/SyncTypeConverter.cs
+++ b/Projects/Dotmim.Sync.Core/SyncTypeConverter.cs
@@ -39,7 +39,8 @@ namespace Dotmim.Sync
                 return Convert.ToUInt64(value);
             else if (typeOfT == typeof(DateTime))
             {
-                if (DateTime.TryParse(value.ToString(), provider, DateTimeStyles.None, out DateTime dateTime))
+                string valueStr = value.ToString();
+                if (DateTime.TryParse(valueStr, provider, DateTimeStyles.None, out DateTime dateTime))
                     return (T)Convert.ChangeType(dateTime, typeOfT, provider);
                 else if (typeOfU == typeof(long))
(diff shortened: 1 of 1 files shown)


Mimetis:
thanks !

andreasfromch:
Thanks a lot @symbiogenesis and @Mimetis 👍 😃!!!

PR #914, opened 2022-12-26
Add npg tests

Mimetis:
Draft PR

Work in progress by @WasimAhmad and @Mimetis


Projects/Dotmim.Sync.PostgreSql/Dotmim.Sync.PostgreSql.csproj
```diff
@@ -1,14 +1,55 @@
...
 	<ItemGroup>
+		<ProjectReference Include="..\Dotmim.Sync.Core\Dotmim.Sync.Core.csproj" />
+		<PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
+	</ItemGroup>
+
+	<ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNet7)'">
+		<PackageReference Include="Npgsql" Version="7.0.1" />
+	</ItemGroup>
+	
+	<ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNet6)'">
```
WasimAhmad, line 42: we need to consider the breaking changes in Npgsql library from 5 to 6 and 6 to 7. specially with datetime

https://www.npgsql.org/doc/release-notes/6.0.html#major-changes-to-timestamp-mapping

WasimAhmad (approved):

WasimAhmad:
can i contribute to this branch directly?

Mimetis:
@WasimAhmad I've added you as Contributor to the repository.
You should be able to Git Pull the branch and contribute
Let me know

WasimAhmad:
> @WasimAhmad I've added you as Contributor to the repository. You should be able to Git Pull the branch and contribute Let me know

its working.

WasimAhmad:
image

As we are using Postgresql functions instead of Procedures
The reason for failing the .net 7 tests is Npgsql db driver breaking change in version => 7.

https://www.npgsql.org/doc/release-notes/7.0.html#a-namecommandtypestoredprocedure-commandtypestoredprocedure-now-invokes-procedures-instead-of-functions.

If i change the commandtype to text then it generate syntex error
image

Mimetis:
We can't use

``` cs
//https://www.npgsql.org/doc/release-notes/7.0.html#a-namecommandtypestoredprocedure-commandtypestoredprocedure-now-invokes-procedures-instead-of-functions.
AppContext.SetSwitch("Npgsql.EnableStoredProcedureCompatMode", true);
//https://www.npgsql.org/doc/release-notes/6.0.html#opting-out-of-the-new-timestamp-mapping-logic
AppContext.SetSwitch("Npgsql.EnableLegacyTimestampBehavior", true);
```

Because, if someone is using `Npgsql` with **stored procedures** in their application, and also uses the `NpgsqlSyncProvider` (to sync) at the same time, it will not work, because we are changing the behavior of the `Npgsql` ADO provider.

I think the best way to be compatible with all versions is to specifically call the functions like this:

``` cs
var connection = new NpgsqlConnection(DBHelper.GetNpgsqlDatabaseConnectionString("tcp_sv_ydysg2vlcfr"));
connection.Open();

// Invoke a function
using var command = new NpgsqlCommand("SELECT * FROM public.\"Address_initialize\"(@sync_min_timestamp, @sync_scope_id)", connection)
{
    Parameters =
    {
        new() {
            ParameterName = "@sync_min_timestamp",
            Value = DBNull.Value
        },
        new() {
            ParameterName = "@sync_scope_id",
            Value = DBNull.Value
        }
    }
};

var reader = command.ExecuteReader();

```

What do you think ?

WasimAhmad:
> We can't use
> 
> ```cs
> //https://www.npgsql.org/doc/release-notes/7.0.html#a-namecommandtypestoredprocedure-commandtypestoredprocedure-now-invokes-procedures-instead-of-functions.
> AppContext.SetSwitch("Npgsql.EnableStoredProcedureCompatMode", true);
> //https://www.npgsql.org/doc/release-notes/6.0.html#opting-out-of-the-new-timestamp-mapping-logic
> AppContext.SetSwitch("Npgsql.EnableLegacyTimestampBehavior", true);
> ```
> 
> Because, if someone is using `Npgsql` with **stored procedures** in their application, and also uses the `NpgsqlSyncProvider` (to sync) at the same time, it will not work, because we are changing the behavior of the `Npgsql` ADO provider.
> 
> I think the best way to be compatible with all versions is to specifically call the functions like this:
> 
> ```cs
> var connection = new NpgsqlConnection(DBHelper.GetNpgsqlDatabaseConnectionString("tcp_sv_ydysg2vlcfr"));
> connection.Open();
> 
> // Invoke a function
> using var command = new NpgsqlCommand("SELECT * FROM public.\"Address_initialize\"(@sync_min_timestamp, @sync_scope_id)", connection)
> {
>     Parameters =
>     {
>         new() {
>             ParameterName = "@sync_min_timestamp",
>             Value = DBNull.Value
>         },
>         new() {
>             ParameterName = "@sync_scope_id",
>             Value = DBNull.Value
>         }
>     }
> };
> 
> var reader = command.ExecuteReader();
> ```
> 
> What do you think ?

Agreed

WasimAhmad:
i'm lost, why these 3 tests are failing?

image

Mimetis:
Reinitialize is failing because when disabling constraints using `ALTER TABLE {0}.{1} DISABLE TRIGGER ALL` will disable constraint (foreign keys) and also triggers (update insert delete) and we don't want that

Mimetis:
`Delete_One_Table` is failing because the function "`<table>_delete`" call is actually deleting **ALL** the rows from the table and not the row with the primary key specified in the arguments.

The strange thing is that it's only failing on **.Net 7** et not on **.Net Core 3.1** ...

Do you have any idea why ?

WasimAhmad:
> I've pushed a commit with a correct version on how to disable constraints without disabling triggers. Let me know if it's correct for you (as i'm not an expert here)

I'm not an expert. I am your student here. :)

WasimAhmad:
> `Delete_One_Table` is failing because the function "`<table>_delete`" call is actually deleting **ALL** the rows from the table and not the row with the primary key specified in the arguments.
> 
> The strange thing is that it's only failing on **.Net 7** et not on **.Net Core 3.1** ...
> 
> Do you have any idea why ?

it failing on all. .Net 3, 6 ,7. I am trying to figure out.

Mimetis:
Yes, indeed !

Mimetis:
Seems we made it :)

image

WasimAhmad:
You made it buddy. I was just try to help. Thanks you. 🙂

WasimAhmad:
I have a suggestion. We should change naming from Npgsql to simple Postgres. 

What did you think?

Mimetis:
Yes, I think it's a good idea.
Honestly I don't know the difference between the naming npgsql and postgres (and postgresql), so I don't have a clear opinion on that

Mimetis:
Another thing I'm thinking about the Postgres provider is to remove completely the functions.
I have a lot of troubles when dealing with migration and modification of the databases.
With functions, every time I need to generate again and again the stored procedures to be sure it's mapping correctly the tables.

If we go for in memory sql statement, it will simplify the use of the provider.
Do you think it's possible to remove the functions, and go for only in memory queries ?

WasimAhmad:
> Yes, I think it's a good idea.
> Honestly I don't know the difference between the naming npgsql and postgres (and postgresql), so I don't have a clear opinion on that

PostgreSQL is the official name of database. And **Npgsql** is the open source . **NET** data provider for **PostgreSQL**

I think we shoud go with official name **PostgreSQL**

Mimetis:
PostgreSqlSyncProvider ?

WasimAhmad:
> PostgreSqlSyncProvider ?

yes

WasimAhmad:
> If we go for in memory sql statement, it will simplify the use of the provider.
> Do you think it's possible to remove the functions, and go for only in memory queries ?

do you want to go with approach like sqlite provider?

WasimAhmad:
What if we extend the provider to support bulk operations like sql provider? 
```
DbStoredProcedureType.BulkTableType
DbStoredProcedureType.BulkUpdateRows
DbStoredProcedureType.BulkDeleteRows
```
If in memory sql statement creates problem with these operations go with functions. 
I personally think it batter to go with functions.

Mimetis:
Yes that's the idea.

I'm currently trying to resolve some issues regarding this "_in memory_" commands to prevent DMS to create functions.
The big problem we are facing is related to the "`_update`" and "`_delete`" commands.
These commands need multiple commands, with inline variables.

All the others are simple `Select` statements.

For "`_update`" and "`_delete`" commands, it's more complicated:
They are using inline variables (like `DECLARE t_update_scope_id uuid;`) 
Inline variables can be used in **PL/pgSQL** **only** in a function or anonymous function using `DO`

We can't do normal **SQL** statement with inline variables (make senses, we are not in a **T-SQL** world here, unfortunately)

So my first test was to try something like this: replace the `Create Function` with a `Do` statement:

``` cs
using var pConnection = new NpgsqlConnection(DBHelper.GetNpgsqlDatabaseConnectionString("tcp_cli_adv"));

using var pCommand = new NpgsqlCommand();
pConnection.Open();

var strCommandString = $@"
            DO $$
            DECLARE t_in_ProductCategoryID varchar (12) = NULL;
            DECLARE ts bigint;
            DECLARE t_update_scope_id uuid;
            BEGIN

            ts := 0;
            SELECT ""ProductCategoryID"", 
            timestamp, update_scope_id FROM public.""ProductCategory_tracking"" 
            WHERE ""ProductCategory_tracking"".""ProductCategoryID"" = @ProductCategoryID LIMIT 1 
            INTO t_in_ProductCategoryID,  ts, t_update_scope_id;

            UPDATE public.""ProductCategory""
            SET  ""ParentProductCategoryId"" = @ParentProductCategoryId
            ,  ""Name"" = @Name
            ,  ""rowguid"" = @rowguid
            ,  ""ModifiedDate"" = @ModifiedDate
            ,  ""Attribute With Space"" = @Attribute_With_Space
            WHERE ""ProductCategoryID"" = @ProductCategoryID
            AND (ts <= @sync_min_timestamp OR ts IS NULL OR t_update_scope_id  = @sync_scope_id OR @sync_force_write = 1);

            GET DIAGNOSTICS @sync_row_count = ROW_COUNT;
            IF (@sync_row_count = 0) THEN
	            INSERT INTO public.""ProductCategory""
	            (""ProductCategoryID"", ""ParentProductCategoryId"", ""Name"", ""rowguid"", ""ModifiedDate"", ""Attribute With Space"")
	            SELECT * FROM ( SELECT @ProductCategoryID, @ParentProductCategoryId, @Name, @rowguid, @ModifiedDate, @Attribute_With_Space) as TMP 
	            WHERE ( t_in_ProductCategoryID IS NULL )
	            OR (ts <= @sync_min_timestamp OR ts IS NULL OR t_update_scope_id = @sync_scope_id OR @sync_force_write = 1)
	            LIMIT 1;
            GET DIAGNOSTICS @sync_row_count = ROW_COUNT;

            END IF;

            IF (@sync_row_count > 0) THEN
	            UPDATE public.""ProductCategory_tracking""
	            SET ""update_scope_id"" = @sync_scope_id, 
		                ""sync_row_is_tombstone"" = FALSE, 
		                ""timestamp"" = (extract(epoch from now())*1000), 
		                ""last_change_datetime"" = now() 
	            WHERE ""ProductCategoryID"" = @ProductCategoryID;
            END IF;
            EXCEPTION WHEN OTHERS THEN
	            GET STACKED DIAGNOSTICS @sync_error_text = MESSAGE_TEXT;
	            @sync_row_count=-1;
            END $$; ";

pCommand.Parameters.Add(new NpgsqlParameter { ParameterName = "@ProductCategoryID", DbType = DbType.String, Size = 12 });
pCommand.Parameters.Add(new NpgsqlParameter { ParameterName = "@ParentProductCategoryID", DbType = DbType.String, Size = 12 });
pCommand.Parameters.Add(new NpgsqlParameter { ParameterName = "@Name", DbType = DbType.String, Size = 150 });
pCommand.Parameters.Add(new NpgsqlParameter { ParameterName = "@rowguid", DbType = DbType.Guid });
pCommand.Parameters.Add(new NpgsqlParameter { ParameterName = "@ModifiedDate", DbType = DbType.Date, });
pCommand.Parameters.Add(new NpgsqlParameter { ParameterName = "@Attribute_With_Space", DbType = DbType.String, Size = 150 });
pCommand.Parameters.Add(new NpgsqlParameter { ParameterName = "@sync_scope_id", DbType = DbType.Guid, });
pCommand.Parameters.Add(new NpgsqlParameter { ParameterName = "@sync_force_write", DbType = DbType.Int64, });
pCommand.Parameters.Add(new NpgsqlParameter { ParameterName = "@sync_min_timestamp", DbType = DbType.Int64 });
pCommand.Parameters.Add(new NpgsqlParameter { ParameterName = "@sync_row_count", DbType = DbType.Int64, Direction = ParameterDirection.Output });
pCommand.Parameters.Add(new NpgsqlParameter { ParameterName = "@sync_error_text", DbType = DbType.String, Direction = ParameterDirection.Output });


pCommand.Parameters["@ProductCategoryID"].Value = "DK_CLOTHES";
pCommand.Parameters["@ParentProductCategoryID"].Value = DBNull.Value;
pCommand.Parameters["@Name"].Value = "DARK CLOTHES";
pCommand.Parameters["@rowguid"].Value = Guid.NewGuid();
pCommand.Parameters["@ModifiedDate"].Value = DateTime.Now;
pCommand.Parameters["@Attribute_With_Space"].Value = "Long clothes";
pCommand.Parameters["@sync_scope_id"].Value = Guid.Empty;
pCommand.Parameters["@sync_force_write"].Value = 0;
pCommand.Parameters["@sync_min_timestamp"].Value = 1772407168003;

pCommand.Connection = pConnection;
pCommand.CommandText = strCommandString;

var reader = await pCommand.ExecuteReaderAsync();
```

Unfortunately, [DO](https://www.postgresql.org/docs/9.0/sql-do.html) statements are **anonymous** functions **without any input parameters** or output results ...
Think it as a `void() => {}` method :)

Fail...

Of course, we can try something without **PL/pgSQL** and go for a fancy **SQL** statement using a `WITH` inline temp table.
Something like that:

``` cs
strCommandString = $@" 
WITH changeset as (
SELECT C.ProductCategoryID, C.ParentProductCategoryID, C.Name, C.rowguid, C.ModifiedDate, C.Attribute_With_Space,
        C.sync_min_timestamp, c.sync_scope_id
FROM (SELECT @ProductCategoryID::varchar(12) as ProductCategoryID, @ParentProductCategoryID::varchar(12) as ParentProductCategoryID,
            @Name::varchar(150) as Name, @rowguid::uuid as rowguid, @ModifiedDate::timestamp without time zone as ModifiedDate,
            @Attribute_With_Space::varchar(150) as Attribute_With_Space, @sync_min_timestamp::bigint as sync_min_timestamp,
            @sync_scope_id::uuid as sync_scope_id) as C
LEFT JOIN public.""ProductCategory_tracking"" as side on side.""ProductCategoryID"" = C.ProductCategoryID
LEFT JOIN public.""ProductCategory"" as base on base.""ProductCategoryID"" = C.ProductCategoryID
WHERE (side.""timestamp"" < C.sync_min_timestamp OR side.""update_scope_id"" = C.sync_scope_id)
OR ( base.""ProductCategoryID"" is null AND (side.""timestamp"" < C.sync_min_timestamp OR side.""timestamp"" IS NULL))
)
                
INSERT INTO public.""ProductCategory""
(""ProductCategoryID"", ""ParentProductCategoryId"", ""Name"", ""rowguid"", ""ModifiedDate"", ""Attribute With Space"")
SELECT ProductCategoryID, ParentProductCategoryID, Name, rowguid, ModifiedDate, Attribute_With_Space
FROM changeset
ON CONFLICT (""ProductCategoryID"") DO
UPDATE SET  ""ParentProductCategoryId"" = EXCLUDED.""ParentProductCategoryId"",  ""Name"" = EXCLUDED.""Name"",  ""rowguid"" = EXCLUDED.""rowguid"",  
        ""ModifiedDate"" = EXCLUDED.""ModifiedDate"",  ""Attribute With Space"" = EXCLUDED.""Attribute With Space"";
";
```

And I was clearly really optimistic with this approach, as it seems to be a good solution to make a `MERGE` statement in one single command.

The problem we have here is **DMS** needs to insert / update a lot of rows:
If one row fails for any reason (like foreign key constraint for example) **DMS** has several options (depending on the user's options):
- Rollback the transaction
- Try the failing row one more time, at the end.
- Try the failing row on the next sync
- Do nothing and continue.

With this simple statement, I tried this:

``` cs
int rowCount;
pConnection.Open();
var pTransaction = pConnection.BeginTransaction();

pCommand.Connection = pConnection;
pCommand.Transaction = pTransaction;
pCommand.CommandText = strCommandString;

try
{
    rowCount = await pCommand.ExecuteNonQueryAsync();
}
catch (Exception ex) { Console.WriteLine(ex.Message); }

try
{
    pCommand.Parameters["@ParentProductCategoryID"].Value = "DDDDDDDD";
    rowCount = await pCommand.ExecuteNonQueryAsync();
}
catch (Exception ex) { Console.WriteLine(ex.Message); }

try
{
    pCommand.Parameters["@ParentProductCategoryID"].Value = DBNull.Value;
    rowCount = await pCommand.ExecuteNonQueryAsync();
}
catch (Exception ex) { Console.WriteLine(ex.Message); }

pTransaction.Commit();
pConnection.Close();
```

As you can see, In the second try catch, i'm forcing a @ParentProductCategoryID to a random value that will generate the foreign key failure.
And I need to catch it, and still continue with the next one.

But unfortunately, I have this error happening:

image

``` bash
current transaction is aborted, commands ignored until end of transaction block
```

I already had this error during the Postgres provider development, using the function.
Luckily, in the function, we can have a "kind of" try catch, like this (preventing the transaction to be aborted)

``` tsql

CREATE OR REPLACE FUNCTION public."ProductCategory_update"(....)
AS $BODY$
   DECLARE ....
BEGIN
   UPDATE .....
EXCEPTION WHEN OTHERS THEN
	GET STACKED DIAGNOSTICS sync_error_text = MESSAGE_TEXT;
	"sync_row_count"=-1;
END; 
$BODY$;
```

My last attempt was to create a temporary function, that will be dropped with the session:

``` cs
var strCommandString = $@"CREATE OR REPLACE FUNCTION pg_temp.""ProductCategory_update""( ....)
```
Then call it:

``` 
var strCommandCall = "Select * from pg_temp.""ProductCategory_update""(@ProductCategoryID ......)
```
The problem here, is that we have to make two calls, and this is not handled by **DMS** (of course we can improve that and make it work, but it will require some work in DMS to achieve this)

Here is my conclusion:

- Using `DO` Statement: Impossible
- Using inline single command, using a `WITH` statement: Possible if we have an "**ALL IN OR NOTHING**" transaction strategy, or impossible
- Using `pg_temp` function: Possible but requires some work in DMS core
- Using classic functions: Already in place, but requires to be cautious with the database (Migration, Provision, Deprovision if the schema is evolving)

Maybe @roji has some advices on that, but in my opinion, if we want to remove the functions, the only way to do it is to use the temporary session functions...

Your thought ?

Mimetis:
> What if we extend the provider to support bulk operations like sql provider?
> 
> ```
> DbStoredProcedureType.BulkTableType
> DbStoredProcedureType.BulkUpdateRows
> DbStoredProcedureType.BulkDeleteRows
> ```

Yes, could be a good improvement for the provider
As we don't have any `TVP` in the **Postgres** world, I guess we can think about `JSON` maybe?

roji:
Hey @Mimetis!

First, about the PostgreSQL vs. Npgsql naming, as noted above Npgsql is just the name of a specific .NET driver for PostgreSQL. If there's ever a different PG driver that you'd want to support, calling your Npgsql-based support "PostgreSQL" would be problematic... So I generally tend towards specifying the driver name in this kind of situation (you never know what the future holds).

If I understand you correctly:

* As an FYI, PostgreSQL 15 introduced MERGE, which works just like SQL Server's; that should allow you to specify logic for when a row already exists/doesn't exist without failing the transaction etc. I'm guessing you want to support older versions of PG, but one possible way here is to support function-less for PG15 via MERGE, and require functions for older versions (the current solutions)?
* If you need to do explicit transaction/savepoint management (e.g. try some operation, retry it if it fails), procedural code is required (PL/pgSQL rather than plain SQL with a CTE/WITH). BTW can you post an example of how your current logic makes this work (with non-temporary functions, or with SQL Server)? Do you make use of PG stored procedures (as opposed to functions, in order to manage transactions)? Do you use savepoints?
* Yeah, DO block indeed don't support parameters. It's fairly typical to just embed literal values in the DO blocks that are sent, but that may not be practical for you (you'd have to know how to generate literals for all supported types, escape strings, etc...).
* Temporary functions are indeed a good way to do this - it's probably how I'd proceed (putting MERGE aside).

Sorry I don't have a magical solution to propose here...

WasimAhmad:
Using pg_temp function seem to be more appropriate for this.

WasimAhmad:
> > What if we extend the provider to support bulk operations like sql provider?
> 
> > 
> 
> > ```
> 
> > DbStoredProcedureType.BulkTableType
> 
> > DbStoredProcedureType.BulkUpdateRows
> 
> > DbStoredProcedureType.BulkDeleteRows
> 
> > ```
> 
> 
> 
> Yes, could be a good improvement for the provider
> 
> As we don't have any `TVP` in the **Postgres** world, I guess we can think about `JSON` maybe?
> 
> 
> 
>
Yes you are right. 
Another option can Composite arrays That is more similar to TVP.

Mimetis:
> So I generally tend towards specifying the driver name in this kind of situation

Indeed, I guess "`NpgsqlSyncProvider`" makes sense. 
Honestly "`PostgreSqlSyncProvider`" or "`NpgsqlSyncProvider`" ... I don't care :D

> As an FYI, PostgreSQL 15 introduced MERGE, I'm guessing you want to support older versions of PG

I'm ok to say that **DMS** is only supporting `PG version > 15`.
Will take a look on the merge statement.

> BTW can you post an example of how your current logic makes this work 

Basically, I'm calling a function called "TABLE_update" that holds the logic of updating a sync row (managing timestamp and so on)

Here is an example of the `ProductCategory_update` function, if you want to have an overview:

<details>

### ProductCategory_update function

``` sql

CREATE OR REPLACE FUNCTION public."ProductCategory_update"(
	"in_ProductCategoryID" character varying DEFAULT NULL::character varying,
	"in_ParentProductCategoryId" character varying DEFAULT NULL::character varying,
	"in_Name" character varying DEFAULT NULL::character varying,
	in_rowguid uuid DEFAULT NULL::uuid,
	"in_ModifiedDate" timestamp without time zone DEFAULT NULL::timestamp without time zone,
	"in_Attribute_With_Space" character varying DEFAULT NULL::character varying,
	sync_scope_id uuid DEFAULT NULL::uuid,
	sync_force_write bigint DEFAULT NULL::bigint,
	sync_min_timestamp bigint DEFAULT NULL::bigint,
	OUT sync_row_count integer,
	OUT sync_error_text text)
    RETURNS record
    LANGUAGE 'plpgsql'
AS $BODY$

DECLARE t_in_ProductCategoryID varchar (12) = NULL;
DECLARE ts bigint;
DECLARE t_update_scope_id uuid;
BEGIN
ts := 0;
SELECT "ProductCategoryID", 
timestamp, update_scope_id FROM public."ProductCategory_tracking" 
WHERE "ProductCategory_tracking"."ProductCategoryID" = "in_ProductCategoryID" LIMIT 1 
INTO t_in_ProductCategoryID,  ts, t_update_scope_id;

UPDATE public."ProductCategory"
SET  "ParentProductCategoryId" = "in_ParentProductCategoryId"
,  "Name" = "in_Name"
,  "rowguid" = "in_rowguid"
,  "ModifiedDate" = "in_ModifiedDate"
,  "Attribute With Space" = "in_Attribute_With_Space"
WHERE "ProductCategoryID" = "in_ProductCategoryID"
AND (ts <= sync_min_timestamp OR ts IS NULL OR t_update_scope_id  = sync_scope_id OR sync_force_write = 1);

GET DIAGNOSTICS "sync_row_count" = ROW_COUNT;
IF (sync_row_count = 0) THEN
	INSERT INTO public."ProductCategory"
	("ProductCategoryID", "ParentProductCategoryId", "Name", "rowguid", "ModifiedDate", "Attribute With Space")
	SELECT * FROM ( SELECT "in_ProductCategoryID", "in_ParentProductCategoryId", "in_Name", "in_rowguid", "in_ModifiedDate", "in_Attribute_With_Space") as TMP 
	WHERE ( t_in_ProductCategoryID IS NULL )
	OR (ts <= sync_min_timestamp OR ts IS NULL OR t_update_scope_id = sync_scope_id OR sync_force_write = 1)
	LIMIT 1;
GET DIAGNOSTICS "sync_row_count" = ROW_COUNT;

END IF;

IF (sync_row_count > 0) THEN
	UPDATE public."ProductCategory_tracking"
	SET "update_scope_id" = sync_scope_id, 
		 "sync_row_is_tombstone" = FALSE, 
		 "timestamp" = (extract(epoch from now())*1000), 
		 "last_change_datetime" = now() 
	WHERE "ProductCategoryID" = "in_ProductCategoryID";
END IF;
EXCEPTION WHEN OTHERS THEN
	GET STACKED DIAGNOSTICS sync_error_text = MESSAGE_TEXT;
	"sync_row_count"=-1;
END; 
$BODY$;
```
</details>

**_Note_**: I think I can improve this function, using a `With` statement or the `Merge` statement. For now, it's just a quick port from the `MySqlSyncProvider`.

Regarding the transaction process, to be sure the transaction will not fail and aborts (from the .net caller), I'm managing the error inside the function itself:

``` sql
CREATE OR REPLACE FUNCTION public."ProductCategory_update"(....)
AS $BODY$
   DECLARE ....
BEGIN
   UPDATE .....
EXCEPTION WHEN OTHERS THEN
	GET STACKED DIAGNOSTICS sync_error_text = MESSAGE_TEXT;
	"sync_row_count"=-1;
END; 
$BODY$;
```
Basically `sync_error_text` is an output parameter containing the error message.
The `try catch` inside the function prevents the function call to fail (and the transaction as well)

In **DMS**, depending on the user decision, once a row is in error, we can:
- Rollback everything
- Try again at the end of the table upserts
- Try again on next sync
- Continue without error

If we are rollbacking everything every time we have an error (old versions of **DMS**), I guess migrating from function to plain SQL statement (using `WITH` or `MERGE`, or both) would be fine.
But as we can't handle correctly the error raised by Postgres and deferring the rollback (or not) of the transaction, I guess is option is not good.

That's being said, I think a viable solution (to remove functions) would be to:
- Migrate to `merge` statement if we consider supporting only Postgres > `v15`.
- Using temporary functions, but will require more works in **DMS** core (as we have a "_pre-phase_" to create the temporary function, before calling update on all rows)

I'm completely biased, but ... omg I miss **T-SQL** here :D

Thanks a ton @roji for the advice.
Let me know if you have some times, one day, to show you the new `NpgsqlSyncProvider` in action ;)

WasimAhmad:
any plans to publish nuget beta?

roji:
Sure thing, I'd be happy to take a look once all this works.

One last note - I don't know everything you're doing, but... If you do go with the MERGE approach, then you may be able to do at least some things without rolling back everything, by avoiding having an error in the first place. In other words, rather than getting a duplicate constraint error (and then having to roll back everything), you can use MERGE to do nothing in that case; you should also be able to get back the IDs of rows which failed (this can be tricky but should be possible), so that you can retry later, ignore or whatever. Depending on what exactly you're trying to support, this may allow you to do at least some things exclusively with MERGE etc.

Mimetis:
@WasimAhmad I'm doing a little bit of refactoring to ensure everything is at the good place, and I'm working on removing the functions (replaced by temporary functions created on the fly) 

Will let you know once I'll push a new beta package

WasimAhmad:
> @WasimAhmad I'm doing a little bit of refactoring to ensure everything is at the good place, and I'm working on removing the functions (replaced by temporary functions created on the fly) 
> 
> 
> 
> Will let you know once I'll push a new beta package

Thanks a lot

Mimetis:
Available, as pre-release :)

https://www.nuget.org/packages/Dotmim.Sync.PostgreSql

image

WasimAhmad:
Congratulations! and thanks a lot. you made a lot of effects to make it work..

WasimAhmad:
Using temp functions can we replace merge with "insert on conflict do update". If it is possible provider will compatible with postgres 9 onward.

PR #929, opened 2023-01-10
Fix inconsistent parameter naming between AddParameter methods

LouisNowell-Nicolle:
This tripped me up briefly when learning filters,
One AddParameter method of the SetupFilter class uses the arguement 'parameterName' which seems correct where as two others refer to it as 'columnName' which isn't accurate when your column and parameter have different names.

### DIFF
--- a/Projects/Dotmim.Sync.Core/Setup/SetupFilter.cs
+++ b/Projects/Dotmim.Sync.Core/Setup/SetupFilter.cs
@@ -65,6 +65,7 @@ namespace Dotmim.Sync
 
         /// <summary>
         /// Add a parameter as input to stored procedure
+        /// <paramref name="parameterName" /> can be anything and later used to compare to a column of the same type with <see cref="AddWhere"/>
         /// For SQL Server, parameter will be added as @{parameterName}
         /// For MySql, parameter will be added as in_{parameterName}
         /// </summary>
@@ -81,6 +82,7 @@ namespace Dotmim.Sync
 
         /// <summary>
         /// Add a parameter based on a column. 
+        /// <paramref name="columnName" /> refers to both the name of the parameter and column
         /// For SQL Server, parameter will be added as @{parameterName}
         /// For MySql, parameter will be added as in_{parameterName}
         /// </summary>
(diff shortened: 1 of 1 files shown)


Mimetis:
Hey @LouisNowell-Nicolle 

Actually, there is two methods for creating a filter parameter:

``` cs
public void AddParameter(string parameterName, DbType type, bool allowNull = false, 
string defaultValue = null, int maxLength = 0) {}
       
public void AddParameter(string columnName, string tableName, string schemaName, bool allowNull = false, 
string defaultValue = null) {}
``` 

The first one, using a `string parameterName` is used to create a _**custom**_ parameter. That's why you have to fill the `DbType`, `allowNull` and `maxLength`, as this parameter is completely customized by you.

On the other hand, when you call the `AddParameter(string columnName ..`, the strategy is completely different.
The parameter **is mapped to a column**, that should exist in one of your tables. That's why you have to fill the `tableName`, to let DMS **knows** where the column is coming from.

As an example, supposing you have a `Name (nvarchar(50) not null)` column in the `ProductCategory` table, you can add a parameter like this:

``` cs
AddParameter("Name", "ProductCategory")
```
**DMS** will retrieve the `Name` column and will infer all the properties for the parameter `@Name`

The other option would be to create your own parameter, that is not mapped to a column, like this

``` cs
AddParameter("ProductNameStartsWith", DbType.String, false);
```

In this example, **DMS** has no idea what is "`ProductNameStartsWith`", and that's why you need to fill the `DbType` and so on...

Make sense ?

LouisNowell-Nicolle:
Ahh thanks I understand now, I wasn't aware there were two distinct types of parameters/strategies and assumed only one. Would it be worth renaming the custom parameter method then to something like `AddCustomParameter` so it's easier to tell they are using different strategies?

Mimetis:
Well, to be honest, I try to have the most consistent methods in **DMS** and the least amount of methods as well...
Adding a new method for adding a parameter seems overengineering to me.
Maybe we can have a better IntelliSense message when calling the method with different arguments, and explain what the difference is between the two?

LouisNowell-Nicolle:
Yep that's a better idea, It was the IntelliSense which tripped me up in the first place anyway. Should I amend this pull request or do you want to make the change?

Mimetis:
As u want :à

LouisNowell-Nicolle:
I've given it a shot but worry I might get some of the details wrong so change as/if you see fit :D

Thanks for creating this tool also, it saved me a lot of time and pain.

PR #934, opened 2023-01-18
Add Limitations and restrictions to overview

VagueGit:
The order of columns matters in DMS. Columns must be in the same order in a table on the client as on the server.

At some point it might be worth having a section for Limitations and Restrictions. For now this puts it up front for DBAs who expect the order of columns does not matter.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -15,6 +15,8 @@ Welcome to Dotmim.Sync
 
 Available for syncing **SQL Server**, **MySQL**, **MariaDB**, **PostgreSQL** and **Sqlite** databases.
 
+Limitations and restrictions: The order of columns matters in DMS. Columns must be in the same order in a table on the client as on the server.
+
 .. note:: The source code is available on `Github <https://www.github.com/mimetis/dotmim.sync>`_. 
    
    This framework is still in beta. There is no support other than me and the time I can put on it. Don't be afraid to reach me out, but expect delay sometimes :)


Mimetis:
You are making a warning for a scenario that is not supported by default.
By default, DMS will creates the client tables. So far, the order is respected.
The other scenario handled is DMS synchronizing to client with an existing database, that has exactly the same schema. Like a backup database restored locally or using Entity Framework to replicate the database structure.

Other thing is that schema is normally retrieved from the server side, but you are doing it on the client side, and this is not a supported scenario, as client side may have different column types (SQLite in particular)

And last thing, I'm not sure AT ALL that making a very edge warning on the index page, where we have the presentation of the framework itself with a trivial example is the right place to do it.

I think you should review your PR accordingly.

Thanks

VagueGit:
There could be many scenarios where there is a pre-existing client database. As one example, a client database that services a local app might be a limited subset of an enterprise server database. The needs of these apps often evolve quickly and erratically ... columns get added when the manager decides they need more info. That column might be added at the end of the table as the dev is not aware of the need to maintain column order for DMS. Enterprise development can be a quick and dirty world.

Another example, a dev may choose to create the local database because DMS does not create the constraints that exist on the server database.

DMS is to be added as a suggestion in documentation for Microsoft Sync Framework #932. Users of Microsoft Sync Framework [Peer-to-Peer Synchronization](https://learn.microsoft.com/en-us/previous-versions/sql/synchronization/sync-framework-2.0/bb726046(v=sql.105)#peer-to-peer-synchronization) are used to schemas that might differ between peer databases. For those devs, this is not an edge scenario. This PR anticipates their needs and expectations.

If you don't want to support those scenarios, then maybe this belongs in a section for Limitations and Restrictions, but I'm afraid I don't know how to do that.

Mimetis:
Remember that this project is OSS, and I spend my spare time on it.

If you think that this scenario is really important, do not hesitate to:
- Clone the project locally
- Modify the code to ensure the scenario is handled correctly
- Test it will all providers, to be sure the scenario is consistent across all databases engine
- Write all the tests
- Write the relevant documentation
- Make the PR

I would love to have it supported, and I will make the PR review.

Thanks

VagueGit:
These scenarios are not currently supported and documentation does not say these scenarios are not currently supported. I've offering to help update the documentation. I'm not in a position to add the missing features.

PR #923, opened 2023-01-04
Fixes issues with tables that contain non-comparable columns

SvdSinner:
Fixes issues with tables that contain non-comparable columns (XML, TXT, NTEXT)

PR #936, opened 2023-01-18
Revert "Fixes issues with tables that contain non-comparable columns"

Mimetis:
Reverts Mimetis/Dotmim.Sync#923 as tests are failing

### DIFF
--- a/Dotmim.Sync.sln.DotSettings
+++ b/Dotmim.Sync.sln.DotSettings
@@ -1,2 +1,0 @@
-﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
-	<s:Boolean x:Key="/Default/UserDictionary/Words/=Noncomparable/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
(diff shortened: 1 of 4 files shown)


SvdSinner:
What do I need to do to fix the tests, or at least see what failed?  I don't have anything failing on my machine.

Mimetis:
Are you able to run the tests locally ?

You must run the tests locally, at least on **SQL Server** and **SQLite** (if you don't have or don't want to use **MySQL** / **Postgres** / **MariaDB**)

You need to run the `Dotmim.Sync.Tests.SqlServerTcpTests`, configured in  `Dotmim.Sync.Tests.Setup.cs`, where you can configure server type and client types:

``` cs
public class SqlServerTcpTests : TcpTests
{
    public SqlServerTcpTests(ITestOutputHelper output, DatabaseServerFixture fixture) : base(output, fixture)
    {
    }

    public override IEnumerable<CoreProvider> GetClientProviders()
    {
        yield return HelperDatabase.GetSyncProvider(ProviderType.Sqlite, "tcp_cli_sqlite_adv");
        yield return HelperDatabase.GetSyncProvider(ProviderType.Sql, "tcp_cli_sql_adv", true);
    }

    public override CoreProvider GetServerProvider() => 
        HelperDatabase.GetSyncProvider(ProviderType.Sql, "tcp_srv_sql_adv", true);
}

```
Here, the **Server** is **Sql Server** and **Clients** are **Sqlite** and **SqlServer** ("`tcp_xxxxx`" are the name of the databases that will be generated on the fly)

Connection strings are retrieved from `appsettings.json`

As you can see, the most straightforward test is "RowsCount" that is not even working here:

image

By the way, all the classic tests for tcp are located in "`/TcpTests/TcpTests.cs`".

Once you have made all the tests passed on Sql Server, you'll need to do the same for `Dotmim.Sync.Tests.SqlServerTcpTests.SqlServerChangeTrackingTcpTests` of course :)

And once it's done, you will have to do the tests for conflicts, http tests and filters ;)

SvdSinner:
Initially, I couldn't get any tests to run, but discovered that the SQL tests required Visual Studio to be run as administrator.  That let me find the bug that was causing problems.  I'll submit a new PR that includes the fix.

PR #938, opened 2023-01-20
Fix non comparable columns

SvdSinner:
Previous bug has been fixed.  Re-submitting.


SvdSinner:
Looks like this pull request only grabbed the fix, rather than grabbing all the code that was in the original pull request. ( https://github.com/Mimetis/Dotmim.Sync/pull/923/ )  You may have to apply them both.  I don't have a lot of experience with re-applying a PR after reverting it.

Mimetis:
Please submit a PR that I can review with all the code and tests passed :)
Will be easier for me!
if you are not sure, the best thing to do is to create a new one from master and reapply your changes.

SvdSinner:
Sounds good.  I will close this PR and am working on the new one.  One question:  Is there an artifact that I need to create and submit for "all tests passed" or just make sure they all pass now that I figured out how to get them to run?

Mimetis:
Just make them pass :)

Issue #910, opened 2022-12-20
Unable to sync Sql Server tables with an XML datatype column


SvdSinner:
When trying to Sync a table with an XML column using the SQL server client, and exception is thrown:
An unhandled exception of type 'Dotmim.Sync.SyncException' occurred in System.Private.CoreLib.dll
The xml data type cannot be selected as DISTINCT because it is not comparable.
├─ [SqlInstance].[DatabaseName].[DefaultScope].[InternalProvisionAsync].Provision:TrackingTable, StoredProcedures, Triggers.Overwrite:False.:The xml data type cannot be selected as DISTINCT because it is not comparable.
├── [SqlInstance].[DatabaseName].[DefaultScope].[InternalCreateStoredProceduresAsync].Table:RenewalRequest.Overwrite:False.:The xml data type cannot be selected as DISTINCT because it is not comparable.
├─── [DefaultScope].[InternalCreateStoredProcedureAsync].Table:RenewalRequest.StoredProcedure:SelectChanges.:The xml data type cannot be selected as DISTINCT because it is not comparable.
└──── The xml data type cannot be selected as DISTINCT because it is not comparable.

I'll look into the source code tomorrow to see if I can find a fix.  If so, I'll post back and send a PR

Mimetis:
Because the `GetChanges` stored procedure (generated) is using a `DISTINCT` in the `SELECT` statement, it won't allow to use **XML** Column.

If you are not explicitly using **XQuery** in your `XML` column, and you are just storing an **xml** value, I would suggest to transform the column into `VARCHAR(MAX)`

SvdSinner:
I do need to support XML columns.  I've been hunting for the code that makes the GetChanges stored procedure,  I'm hoping I can change the query when XML columns are present.

SvdSinner:
It looks like I can add support for non-comparable columns (xml, text, ntext and maybe more if needed).  I've got the code written and it appears to work in very early testing.
Unfortunately, I'm at the end of my day on the last day I'm working this year, so I probably won't get to fully test it or submit it as a pull request until next year.

SvdSinner:
Pull request submitted.  Be aware that I have only tested it on the databases that I care about so additional testing would be prudent.  Also, I am not currently using change tracking so I couldn't test that chunk.  If you have any questions, let me know.

SvdSinner:
https://github.com/Mimetis/Dotmim.Sync/pull/941 submitted to replace previous pull requests

[...]

PR #941, opened 2023-01-23
All code to support non-comparable columns

SvdSinner:
This replaces https://github.com/Mimetis/Dotmim.Sync/pull/938 and https://github.com/Mimetis/Dotmim.Sync/pull/923.

Fixes https://github.com/Mimetis/Dotmim.Sync/issues/910

### DIFF
--- a/Projects/Dotmim.Sync.SqlServer.ChangeTracking/Builders/SqlChangeTrackingBuilderProcedure.cs
+++ b/Projects/Dotmim.Sync.SqlServer.ChangeTracking/Builders/SqlChangeTrackingBuilderProcedure.cs
@@ -2,11 +2,11 @@
 using Dotmim.Sync.SqlServer.Builders;
 using Dotmim.Sync.SqlServer.Manager;
 using Microsoft.Data.SqlClient;
+using System;
 using System.Data;
 using System.Data.Common;
 using System.Linq;
 using System.Text;
-using static Azure.Core.HttpHeader;
 
 namespace Dotmim.Sync.SqlServer.ChangeTracking.Builders
 {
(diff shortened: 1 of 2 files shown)


Mimetis:
The problem I see here is that all tests passed because the Adventureworks generated database has no `xml` or `ntext` column actually.

You can change that by specifying that one column from one table (that is filtered) has a `ntext` column instead of `varchar()`

Like for example `SalesOrderHeader` has a column `Comment`, that you can use for that in [AdventureWorksContext.cs](https://github.com/SvdSinner/Dotmim.Sync/blob/FixNonComaparableColumnsFixed/Tests/Dotmim.Sync.Tests/Models/AdventureWorksContext.cs), you can add this property (around line 600):

```  cs
if (this.ProviderType == ProviderType.Sql)
    entity.Property(e => e.Comment).HasColumnType("ntext");

```
That way, the `Comment` field will be generated as a `ntext` column:

image

Once you have modified the database generation, you will see that errors are showing up ;)

SvdSinner:
I forgot to mention that I had found some bugs that I'm working out.  I tried using DMS on a 15 yr old database that was designed by people who had no idea what they were doing.  It has everything ugly in it:  Text, NText, Image columns, Tables with no primary keys, column names with special characters and very little referential integrity.  

It is a dumpster-fire of a database, but it's been great for testing and finding edge cases.  Once I work threw all the issues, I will make a commit  with the fixes for everything I have found.

SvdSinner:
I think this is ready for merging.  There are still a few failing tests (below) that I do not think are related to anything in this PR.

Still failing (Net 6.0) tests:
SqlServerChangeTrackingHttpFilterTests.DeleteTwoRowsInTwoTablesOnClientSide
SqlServerTcpFilterTests.DeleteTwoRowsInTwoTablesOnClientSide
SqlServerUnitRemoteOrchestratprTests.RemoteOrchestrator_HttpGetChanges_WithFilters_ShouldReturnDeletedRowsCount
SqlServerUnitRemoteOrchestratorTests.RemoteOrchestrator_HttpGetEstimatedChanges_WithFilters_ShouldReturnDeletedRowsCount

SvdSinner:
I've merged in all recent changes and re-ran all the tests.  There is now only one failing test (Net 6.0):
SqlServerTcpFilterTests.DeleteTwoRowsInTwoTablesOnClientSide
Again, I don't think this PR has anything at all to do with that test, so I don't believe the failure is relevant.

RaniRishmawi:
Hello team, first I would like to thank all the maintainers for a great library. however, I am facing the issue of ntext type not being supported. I believe this issue has to do with it. Is this to be released soon? Is there anything that I can help with to move this fix to Nuget packages?

Mimetis:
`ntext`, `text` and `image` will be removed in the future versions of SQL Server. (https://learn.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?view=sql-server-ver16)

This PR adds the support of these types, but with a lot of complexicity.
It's a great feature, but I believe we should not add it tot the next release of DMS, for multiple reasons.

I believe we should encourage people to migrate to `nvarchar(max)` or `varbinary(max)` instead of using these old, non performant, types.
The support of these types in **DMS** adds a lot of complexity with some tradeoffs.

I think it's a good example, as a provider, but should not be add to the core source of `DMS` (especially in `SqlSyncProvider`)

Issue #962, opened 2023-02-09
Sync fails with duplicate key constraint failue


Craige:
Syncing from Sqlite to MySQL. First sync syncs to from the  server to the client. Subsequent sync (to sync lo local change to the remote) fails with:

```
[InternalApplyThenGetChangesAsync]..[InternalApplyThenGetChangesAsync]..[InternalSaveScopeInfoClientAsync]..Duplicate entry '3ea49ef3-02c8-4cd0-af39-7acffab1338f-v1-YRV8q709t1f3m4tvB/5mP...' for key 'PRIMARY' (Dotmim.Sync.SyncException)
   at Dotmim.Sync.RemoteOrchestrator.InternalApplyThenGetChangesAsync(ScopeInfoClient cScopeInfoClient, ScopeInfo cScopeInfo, SyncContext context, ClientSyncChanges clientChanges, DbConnection connection, DbTransaction transaction, CancellationToken cancellationToken, IProgress`1 progress)
   at Dotmim.Sync.SyncAgent.SynchronizeAsync(String scopeName, SyncSetup setup, SyncType syncType, SyncParameters parameters, CancellationToken cancellationToken, IProgress`1 progress)
   at Dotmim.Sync.SyncAgent.SynchronizeAsync(String scopeName, SyncSetup setup, SyncType syncType, SyncParameters parameters, CancellationToken cancellationToken, IProgress`1 progress)
[...]
```
It does appear before this that the client change is synced to the server, but something is going wrong here. This is with a minimal stock sync setup:

```
public class Foo
{
    public MariaDBSyncProvider ServerProvider { get; set; }
    public SqliteSyncProvider ClientProvider { get; set; }
    public SyncSetup Setup { get; set; }
    public SyncAgent Agent { get; set; }

    public Foo()
    {       
        ServerProvider = new MariaDBSyncProvider("Data Source=localhost,3306;Initial Catalog=project;User ID=user;Password=password");
        var path = ""; // redacted for simplicity
        ClientProvider = new SqliteSyncProvider(path);

        Setup = new SyncSetup("Table1", "Table2");

        Agent = new SyncAgent(ClientProvider, ServerProvider);
    }
    public async Task onEvent()
    {
        // Persist local changes...

        //...

        // Sync to remote...
        var result = await Agent.SynchronizeAsync("v1", Setup, SyncType.ReinitializeWithUpload);
        Console.WriteLine(result);
    }
}

```
Using:
```
[...]

PR #969, opened 2023-02-19
Set sync_scope_id parameter as Guid consistently in Buider classes.

Craige:
Mixing `sync_scope_id` parameter types can cause issues with syncing. It seems in certain scenarios, GUID parameter converted as a string results in a malformed GUID in the form of `a0a0a0a0a-0001-0000-0000-000000000000` as seen in #962.

GUID values should always be always be converted as `DbType.Guid` consistently.


Craige:
Glad to help. 

Thanks for the merge!

PR #1011, opened 2023-05-08
Vectorize graphics

symbiogenesis:
This should make the README on github look smooth and sharp. The Architecture01 bitmap was noticeably poor quality when viewed in dark mode, especially.

This should improve the way the Architecture01 graphic appears in dark mode, and will be infinitely scalable, and easier to modify going forward.

You can see what the new vectors will look like [here](https://github.com/symbiogenesis/Dotmim.Sync/tree/vectorize-graphics)


Mimetis:
Wow ! Did you re-created everything from scratch ?

symbiogenesis:
> Wow ! Did you re-created everything from scratch ?

I used the open source Inkscape program to trace the bitmap, which was able to scrape the logo and the clipart, with some minor path editing, and then I just quickly did a bit of editing using the shape tools to generate the arrows, and added my own text labels on top of the old ones, then deleted the layers below and coverted the font objects to paths to ensure cross-platform consistency.

There's also free libraries of SVG clipart online that one could've used to re-create a lot of this.

PR #1012, opened 2023-05-08
Specify StringComparison directly

symbiogenesis:
This should be slightly faster and also more explicit. It is the recommended pattern in StyleCop. I used an automated refactoring which ensures that the StringComparison type was unchanged.

see: [CA1307](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1307), [CA1310](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1310)

### DIFF
--- a/Projects/Dotmim.Sync.Core/Orchestrators/Upgrade/LocalOrchestrator.Upgrade.cs
+++ b/Projects/Dotmim.Sync.Core/Orchestrators/Upgrade/LocalOrchestrator.Upgrade.cs
@@ -466,11 +466,11 @@ namespace Dotmim.Sync
                                 throw new Exception("Your setup to migrate contains one or more filters. Please use a SyncParameters argument when calling ManualUpgradeWithFiltersParameterAsync() (or SynchronizeAsync()).");
 
                             foreach (var setupParameter in setupParameters)
-                                if (!parameters.Any(p => p.Name.ToLowerInvariant() == setupParameter.ToLowerInvariant()))
+                                if (!parameters.Any(p => string.Equals(p.Name, setupParameter, SyncGlobalization.DataSourceStringComparison)))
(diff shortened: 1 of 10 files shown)


Mimetis:
Thanks for this small refactoring.
We have a static class called `SyncGlobalization` that is responsible of setting the `StringComparison` property
(And it should be used everywhere :D )

I think this **PR** should uses this class instead of setting it everywhere.

Another point I'm considering right now as I see this, is maybe to remove the `SyncGlobalization` static class and set a new property in the `SyncOptions` class, that can be set by the user himself, if needed

What do you think ?

symbiogenesis:
I think that these cases would ideally be determined on a case by case basis. For instance, it is possible that certain comparisons are only checking for a given SQL keywords that are defined in the ANSI/ISO specification, and could be set to the fastest option of Ordinal (or OrdinalIgnoreCase), or you need to check some user data at runtime for equality down to the bit level for whatever reason. But other times, you would want to be more forgiving.

Having the assumptions be explicit helps at least make it more obvious that this is what is happening.  Thus, this becomes a stepwise improvement in any case.

CA1307, CA1310 and so on could be enforced in the build pipeline going forward with EditorConfig, which I recommend adding anyways.

```
# Code analyzers
dotnet_diagnostic.CA1307.severity = error
dotnet_diagnostic.CA1310.severity = error
```

Defining a specific CultureInfo might be a good idea for SyncOptions, and we could use that to set the culture without the user needing to override the global current CultureInfo property within .NET.

But StringComparison is less likely to be something that we need to define as part of SyncOptions. Same with StringComparer.

symbiogenesis:
I updated this to use SyncGlobalization because the end result is identical, and it does ensure consistency.

symbiogenesis:
Fixed build on netstandard2.0

PR #1013, opened 2023-05-08
Make some methods and classes static

symbiogenesis:
May help with performance and also help to enforce correctness. Accomplished with automatic refactoring via [Roslyn Analyzers](https://github.com/dotnet/roslyn-analyzers).

see [CA1822](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1822)

### DIFF
--- a/Projects/Dotmim.Sync.Core/Logger/SyncLogger.cs
+++ b/Projects/Dotmim.Sync.Core/Logger/SyncLogger.cs
@@ -106,7 +106,7 @@ namespace Dotmim.Sync
         /// <summary>
         /// Write a messages without returning to new line
         /// </summary>
-        internal void Write(OutputWriter outputWriter, string message, ConsoleColor? background = default, ConsoleColor? foreground = default)
+        internal static void Write(OutputWriter outputWriter, string message, ConsoleColor? background = default, ConsoleColor? foreground = default)
         {
             outputWriter.SetColor(background, foreground);
             outputWriter.Write(message);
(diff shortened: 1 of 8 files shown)


Mimetis:
I don't buy it. Well, not all :)

A lot of methods, even if not depending on the class itself, are following a "pattern"
`Internal***` methods belong to the instance of the class, with specific public methods exposed.

Indeed, sometimes some of them "_can_" be static, but depending on future iterations / behaviors or new features, we may need some class properties.

That being said, some of them makes a lot of sense, like `SyncLogger` or `SyncTypeConverter` where static methods will be a considerable perf upgrade

For all the providers the benefits is not important as the methods are called once during provisioning, and never during a sync process.

symbiogenesis:
> I don't buy it. Well, not all :)
> 
> A lot of methods, even if not depending on the class itself, are following a "pattern" `Internal***` methods belong to the instance of the class, with specific public methods exposed.
> 
> Indeed, sometimes some of them "_can_" be static, but depending on future iterations / behaviors or new features, we may need some class properties.
> 
> That being said, some of them makes a lot of sense, like `SyncLogger` or `SyncTypeConverter` where static methods will be a considerable perf upgrade
> 
> For all the providers the benefits is not important as the methods are called once during provisioning, and never during a sync process.

I updated the PR to only make a smaller number of more impactful and less controversial changes.

Mimetis:
Seems way better now :)

PR #1018, opened 2023-05-08
Optimize StringBuilder usage

symbiogenesis:
This is another micro-optimization. I accomplished it with the help of automatic refactoring via [Roslynator](https://github.com/JosefPihrt/Roslynator)


Mimetis:
This optimization adds too much complexity in code reading.
It's not worth it, IMO

symbiogenesis:
fair enough

PR #1019, opened 2023-05-08
Use faster approach for generating random names in tests

symbiogenesis:
I noticed that a lot of the tests are using a particular scheme to randomize names. They also have a long wait in the case of a collision.

This new approach will be very very unlikely to collide, and thus the tests should also run faster, because it will almost never have to wait a full second (Thread.Sleep) because of a collision.


symbiogenesis:
Even if it doesn't fix the tests, the new approach is probably faster. So it could be worth merging.

symbiogenesis:
Was hoping this would fix the failing tests, because I saw them pass locally, but the same conflict tests are still failing in the pipeline. But this could still be a performance increase.

Mimetis:
We have some failing tests for `SqlChangeTrackingProvider`, but that's normal.
Tests are "testing" something that is not handled by this provider.
I just need to make an exception and remove them when testing the `SqlChangeTrackingProvider`.

PR #1022, opened 2023-05-08
Newtonsoft JsonSerializer is thread-safe

symbiogenesis:
The consensus seems to be that JsonSerializer is thread-safe

https://stackoverflow.com/questions/36186276/is-the-json-net-jsonserializer-threadsafe


symbiogenesis:
Looks like we could probably remove the JsonSerializerFactory on the server side, as well.

PR #1024, opened 2023-05-09
Simplify LocalJsonSerializer

symbiogenesis:
This was a fairly complicated class, and it should be easier to read now. The commits should be self-explanatory, when read one by one.

### DIFF
--- a/Projects/Dotmim.Sync.Core/Serialization/LocalJsonSerializer.cs
+++ b/Projects/Dotmim.Sync.Core/Serialization/LocalJsonSerializer.cs
@@ -378,15 +378,7 @@ namespace Dotmim.Sync.Serialization
 
                                 if (array == null || array.Length < 2)
                                 {
-                                    string rowStr = "";
-
-                                    if (array != null)
-                                    {
-                                        rowStr = "[";
-                                        for (var o = 0; o < array.Length; o++)
-                                            rowStr += array[o].ToString() + ",";
-                                        rowStr += "]";
-                                    }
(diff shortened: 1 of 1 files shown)


Projects/Dotmim.Sync.Core/Serialization/LocalJsonSerializer.cs
```diff
@@ -475,12 +426,12 @@ private static SyncTable GetSchemaTableFromReader(JsonTextReader reader, JsonSer
 
             schemaTable = new SyncTable(tableName, schemaName);
 
-            for (int i = 0; i < includedColumns.Count; i++)
+            foreach (var includedColumn in includedColumns)
```
workgroupengineering, line 429: Please consider reverting this change for [this](https://www.meziantou.net/fastest-way-to-enumerate-a-list-t.htm) reason.

symbiogenesis: I also liked that this approach ensures the array lookup by index happens only once instead of 3 times.

Perhaps the compiler optimizes that away, or we could create a local variable.

Mimetis:
We have a lot of errors regarding this PR : https://dev.azure.com/dotmim/Dotmim.Sync/_build/results?buildId=1446&view=results

Projects/Dotmim.Sync.Core/Serialization/LocalJsonSerializer.cs
workgroupengineering: you have same result with `includedColumn  = includedColumns[i]`

PR #1030, opened 2023-05-11
Cache all DbCommands generated by adapters

symbiogenesis:
This uses a similar strategy to what was already used in the SqlSyncAdapter in the SQL Server project, with respect to the derived parameters cache.

The cache key includes the database name and the other two parameters of the GetCommand() function, and thus it is a simple case of memoization and is unlikely to fail.

Because the ConcurrentDictionary was made static, just like the one for the derived parameters, it needed to follow the notice in the comment there, that the key should include the database name. Which is why I added the DbConnection parameter to GetCommand()

### DIFF
--- a/Projects/Dotmim.Sync.Core/DbSyncAdapter.cs
+++ b/Projects/Dotmim.Sync.Core/DbSyncAdapter.cs
@@ -1,4 +1,4 @@
-using Dotmim.Sync.Builders;
+﻿using Dotmim.Sync.Builders;
 using System;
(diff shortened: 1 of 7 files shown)


symbiogenesis:
The only reason I can see that would make this fail is on the SQL Server adapter if someone manually set the SqlObjectNames or SqlMetadata properties after already initializing the adapter.

This seems like a bad idea, and in my opinion we should set those two properties to have only private setters.

Issue #1141, opened 2024-02-29
SynchronousProgress<T> in WinUI 3/AppSdk app: The send method is not supported, use Post instead.


Christian-Fenzl:
If SynchronousProgress<T> is used in a WinUI 3/AppSdk application an exception is thrown when the sync is started which tells:
"The send method is not supported, use Post instead."

Changing line 73 in /Projects/Dotmim.Sync.Core/Interceptors/SynchronousProgress.cs 
`m_synchronizationContext.Send(m_invokeHandlers, value);`
to
`m_synchronizationContext.Post(m_invokeHandlers, value);`
and everything works fine.

Thanks for the great lib!

Mimetis:
`SynchronousProgress` should only be used in a console application.

Use a classich `Progress<T>` instead.

That's being said, if `Post` is better than `Send` it's worth !

Can you make a quick PR ?

PR #1145, opened 2024-03-14
#1141 : SynchronousProgress<T> in WinUI 3/AppSdk app

Christian-Fenzl:
See #1141 

If SynchronousProgress is used in a WinUI 3/AppSdk application an exception is thrown when the sync is started which tells:
"The send method is not supported, use Post instead."

Changing line 73 in /Projects/Dotmim.Sync.Core/Interceptors/SynchronousProgress.cs
m_synchronizationContext.Send(m_invokeHandlers, value);
to
m_synchronizationContext.Post(m_invokeHandlers, value);
and everything works fine.


Mimetis:
Thanks, good catch !
That being said, the `SynchronousProgress<T>` is only meant to be used in console application.
For **WinUI** / **UWP** and so on, I think `Progress<T>` is the way to go, in my opinion.

PR #1150, opened 2024-04-07
use static JsonSerializer

symbiogenesis:
Going along with my other PR https://github.com/Mimetis/Dotmim.Sync/pull/1022 we can safely assume that JsonSerializer is thread-safe, as per the advice of the library author James Newton-King.

A static serializer means we can avoid repeated object creation overhead

JsonSerializer is preferred over JObject because it is faster when we know the type ahead of time. Which we do.

This removes the SerializeAsync(object) overload because the generic version handles that perfectly.

### DIFF
--- a/Projects/Dotmim.Sync.Core/Serialization/ISerializer.cs
+++ b/Projects/Dotmim.Sync.Core/Serialization/ISerializer.cs
@@ -26,10 +26,6 @@ namespace Dotmim.Sync.Serialization
         Task<object> DeserializeAsync(Stream ms, Type type);
         Task<T> DeserializeAsync<T>(Stream ms);
 
-        Task<byte[]> SerializeAsync(object obj);
         Task<byte[]> SerializeAsync<T>(T obj);
     }
-
-
-
 }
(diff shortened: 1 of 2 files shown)


Mimetis:
make sense, thanks !

Mimetis:
hey @symbiogenesis 
I've merged your PR, but I'm a little bit concerned about the "_non async_" serialization / deserialization of this new static `JsonObjectSerializer`.

I didn't do any tests so far, but can we have potential blocking UI issues here as we are not async when serialize / deserialize ?

symbiogenesis:
If tests look bad then we could revert, although I don't see an issue right now.

The long-term best solution is to update to System.Text.Json and I was hoping to eventually make a PR to do that as well. That would be fully async and better performance.

I will see if I can make that happen sooner.

Mimetis:
Agree on updating to `System.Text.Json`
It's on my todo list, but if you want to make a PR, you are more than welcome !

Only thing to keep in mind is to maintain compatibility with `.Net Standard 2.0` (if we want to keep this compatibility #1151 )

PR #1151, opened 2024-04-07
add .NET 8 support, and remove .NET 7 & .NET Core 3.1

symbiogenesis:
This adds support for .NET 8

Removed .NET 6 support because it will be end of life in 6 months

**EDIT: Keeping .NET 6, and removing .NET 7 and .NET Core 3.1**

### DIFF
--- a/Projects/Dotmim.Sync.Core/Dotmim.Sync.Core.csproj
+++ b/Projects/Dotmim.Sync.Core/Dotmim.Sync.Core.csproj
@@ -2,10 +2,9 @@
 	<PropertyGroup>
 		<TargetFrameworkNetStandard>netstandard2.0</TargetFrameworkNetStandard>
 		<TargetFrameworkNet6>net6.0</TargetFrameworkNet6>
-		<TargetFrameworkNet7>net7.0</TargetFrameworkNet7>
-		<TargetFrameworkNetCore>netcoreapp3.1</TargetFrameworkNetCore>
+		<TargetFrameworkNet8>net8.0</TargetFrameworkNet8>
 
-		<TargetFrameworks>$(TargetFrameworkNetStandard);$(TargetFrameworkNetCore);$(TargetFrameworkNet6);$(TargetFrameworkNet7)</TargetFrameworks>
+		<TargetFrameworks>$(TargetFrameworkNetStandard);$(TargetFrameworkNet6);$(TargetFrameworkNet8)</TargetFrameworks>
 		<Authors>Sébastien Pertus</Authors>
 		<Company>Microsoft</Company>
 		<RootNamespace>Dotmim.Sync</RootNamespace>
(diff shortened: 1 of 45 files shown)


vunder:
What's the point of removing LTS net6 which has 6 more months of the support, but keeping net7 that has end support at May 14

Mimetis:
btw, should we consider to remove support for `.Net core 3.1` and `.Net standard 2.0` ?

If we are removing unsupported versions, we should have only `.Net 6` and `.Net 8`

Is it making sense ?

symbiogenesis:
.NET Framework 4.8 has a very long supported life left. Which means keeping netstandard 2.0 makes sense.

So maybe netstandard 2.0, .NET 6, and .NET 8 are the ideal ones to have.

Which means removing .NET 7 and .NET Core 3.1

symbiogenesis:
Removing .NET Standard 2.0 support, however, would free up the adoption of the latest libraries in general. And reduce the need for a lot of conditional #if blocks and special property groups defining different library versions for different runtimes.

Also having less pipelines would make the pipelines faster, and everything could be more maintainable.

The main library I want to switch to now, System.Text.Json, supports .NET Standard 2.0 so either choice seems fine to me.

Mimetis:
> The main library I want to switch to now, System.Text.Json, supports .NET Standard 2.0 so either choice seems fine to me.

Are you sure ? :)

image

Mimetis:
Hum, ok, the doc is more explicit:

We can have it, if we are installing it, through a nuget package : https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/overview#how-to-get-the-library

VagueGit:
> btw, should we consider to remove support for `.Net core 3.1` and `.Net standard 2.0` ?

No! We have hundreds of companies syncing data through a UWP app. We need .Net standard 2.0 support.

symbiogenesis:
Updated to target netstandard 2.0, .NET 6, and .NET 8

Mimetis:
Thanks @symbiogenesis !

It's cool to have the last `.NET Framework` version right now, but it seems you removed .. too much .. in the tests :)
How are we testing the `.NET Standard 2.0` version ?

symbiogenesis:
Good point. Restored the .NET Core 3.1 pipeline.

Mimetis:
Almost good, but you are missing the .Net Core 3.1 framework in https://github.com/Mimetis/Dotmim.Sync/blob/bcdb9e98e2ba70f396a49c34271fd88a1135b04d/Tests/Dotmim.Sync.Tests/Dotmim.Sync.Tests.csproj

And I don't know why, but seems we have some errors in the tests.

symbiogenesis:
> Almost good, but you are missing the .Net Core 3.1 framework in https://github.com/Mimetis/Dotmim.Sync/blob/bcdb9e98e2ba70f396a49c34271fd88a1135b04d/Tests/Dotmim.Sync.Tests/Dotmim.Sync.Tests.csproj
> 
> And I don't know why, but seems we have some errors in the tests.

Pushed what I hope is a fix for those errors.

There was no netcoreapp3.1 target in the tests as they exist currently in the master branch. So this PR is not a regression, at minimum.

Mimetis:
> There was no netcoreapp3.1 target in the tests as they exist currently in the master branch. So this PR is not a regression, at minimum.

AH AH you're right, something I'll fix later !

PR #1153, opened 2024-04-07
show root-level files in solution

symbiogenesis:
This is a convenience that allows easily modifying root-level files inside of Visual Studio


Mimetis:
love it !

Issue #1165, opened 2024-04-10
Turn on nuget lockfiles and LockedMode


symbiogenesis:
Nuget provides a lockfiles option similar to npm. Turning it on and forcing the compiler to respect it can prevent supply chain attacks like the infamous leftpad attack.

https://www.meziantou.net/faster-and-safer-nuget-restore-using-source-mapping-and-lock-files.htm

symbiogenesis:
@Mimetis hopefully this one makes sense. It is a newer feature of nuget, but very useful from a security and determinism standpoint.

Issue #1162, opened 2024-04-10
Speed up tests


symbiogenesis:
Lots of tests have identical setup and could make use of [shared context](https://xunit.net/docs/shared-context) in order to speed things up.

Relatedly, we could look at improving parallelism, such as making more separation between classes when there are different test setups. Tests within classes run sequentially, but the different classes are run in parallel.

There are other possibilities, but this seems to be the safest and lowest hanging fruit.

Mimetis:
I have already used shared context, but it's not working.
You can't use the same db for each test, coz you never know the state of the database, the records it contains, and even the change in the schema

If you have any idea, do not hesitate to share them here

PR #1168, opened 2024-04-11
Speedup pipelines

### LINKED ISSUE #1167: Speed up pipelines
The pipelines could be optimized in some ways:

1) the SDKs can be downloaded only as needed, instead of downloading all 3 on every pipeline.

2) Nuget packages can be cached between builds instead of needing to be re-downloaded each time

3) Docker containers could be optimized so that only the correct docker container needed for each test is run

4) dotnet CLI usage could be optimized to prevent duplicate steps.

symbiogenesis:
This may go some ways towards speeding up the pipelines

I think it will save a couple minutes per pipeline. Refactoring the tests, as described in #1162 would help more substantially.

The cache step won't save any time on the first run, but subsequent runs will benefit.

Generating lockfiles was necessary in order to create a perfect cache key

Fixes #1167 and #1165


symbiogenesis:
@Mimetis This should fix the pipelines and also improve them a bunch.

And it closes two of the issues I just opened, with some explanation.

symbiogenesis:
@Mimetis this should make the pipelines green. I just resolved the conflicts from today.

Mimetis:
@symbiogenesis  nice, I'll look this week end for sure

Mimetis:
Hey @symbiogenesis , I'm struggling a little bit with the tests pipelines since this PR

Seems the tests are not run when targeting .net core 3.1
Seems the runner is not able to find the related tests:

```
No test matches the given testcase filter `Dotmim.Sync.Tests.SqlServerTcp` in 
/home/vsts/work/1/s/Tests/Dotmim.Sync.Tests/bin/Debug/netcoreapp3.1/Dotmim.Sync.Tests.dll
```

https://dev.azure.com/dotmim/Dotmim.Sync/_build/results?buildId=1631&view=logs&j=196c4937-6990-58b5-7d18-234ac2f0888b&t=45bab8b5-ece4-5e34-d7da-aaa6e8c8056a

Any idea ?

Mimetis:
hum I don't know what I've done, but it's working now :)

PR #1195, opened 2024-06-30
fix test

symbiogenesis:
We can see from the referenced code below that the correct data type for the 9th column is `DateTime`

```csharp
public static SyncTable GetSimpleSyncTable(int rowsCount = 1)
{
    SyncTable tCustomer = new SyncTable("Customer");
    tCustomer.Columns.Add(new SyncColumn("ID", typeof(Guid)));
    tCustomer.Columns.Add(new SyncColumn("Name", typeof(string)));
    tCustomer.Columns.Add(new SyncColumn("Column_Int16", typeof(short)));
    tCustomer.Columns.Add(new SyncColumn("Column_Int32", typeof(int)));
    tCustomer.Columns.Add(new SyncColumn("Column_Int64", typeof(long)));
    tCustomer.Columns.Add(new SyncColumn("Column_UInt16", typeof(ushort)));
    tCustomer.Columns.Add(new SyncColumn("Column_UInt32", typeof(uint)));
    tCustomer.Columns.Add(new SyncColumn("Column_UInt64", typeof(ulong)));
    tCustomer.Columns.Add(new SyncColumn("Column_DateTime", typeof(DateTime)));
    tCustomer.Columns.Add(new SyncColumn("Column_DateTimeOffset", typeof(DateTimeOffset)));
```

The rest of the types in the test aren't all matching, though. Should they be?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -274,3 +274,6 @@ __pycache__/
 /.vscode
 
 /docs/_build
+
+# InferSharp config file
+.infersharpconfig
--- a/Projects/Dotmim.Sync.Core/Serialization/ObjectToInferredTypesConverter.cs
+++ b/Projects/Dotmim.Sync.Core/Serialization/ObjectToInferredTypesConverter.cs
@@ -17,7 +17,7 @@ namespace Dotmim.Sync.Serialization
                 JsonTokenType.False => false,
                 JsonTokenType.Number when reader.TryGetInt64(out long l) => l,
                 JsonTokenType.Number => reader.GetDouble(),
-                JsonTokenType.String when reader.TryGetDateTime(out DateTime datetime) => datetime,
+                JsonTokenType.String when reader.TryGetDateTimeOffset(out DateTimeOffset dto) => dto,
                 JsonTokenType.String => reader.GetString()!,
                 _ => JsonDocument.ParseValue(ref reader).RootElement.Clone()
             };


Mimetis:
The thing is that when you have a number in a **JSON** payload, you can't know if it's `byte`, `short`, `int`, `long` ...
Except if we decide to add the type in the **JSON** node 

That's why you have `long` as the default wherever it can apply:

- Same for everything serialized to string (`String`, `Guid`, `Timespan`, `Char`) ...
- Same for `DateTime`, `DateTimeOffset` and `DateOnly` (that we don't support yet). The best deal is to convert all date to the most precise : `DateTimeOffset`

That's why `DateTime`, and `DateTimeOffset` fields will be deserialized to `DateTimeOffset` only.
At least this is the role of the base serializer (`JsonObjectSerializer`) that has no clue about the types (and that's fine)

The second layer is the `LocalJsonSerializer` object (that is using the `JsonObjectSerializer`).
Since `LocalJsonSerializer` **HAS** the table schema, it can easily **Convert Back** to the desired type anything outcoming from the `JsonObjectSerializer.DeserialiseAsync()` method:

Look the `GetRowsFromFile()` method, line `~~450` :

``` cs
if (value != null)
{
    var convertedValue = SyncTypeConverter.TryConvertTo(value, columnType);
    values[index] = convertedValue;
}
```

This converter will transform (for example) a column with type `int`, from `long` to `int`.
Same for any `DateTimeOffset` property that should be convert back to `DateTime`.

Make Sense ?

The only concern I have right now is that we have multiple redundancy in the serialization.
A lot of classes should be refactored to use only one single method to serialize deserialize.
We have multiple times, the same code, that should be put in an extension method (on `Utf8JsonReader` or `JsonReader`), I guess.

`ArrayJsonConverter.cs` :

https://github.com/Mimetis/Dotmim.Sync/blob/e8948f7983901550b14647eedd5b2e11f04a290b/Projects/Dotmim.Sync.Core/Serialization/ArrayJsonConverter.cs#L26-L40

`ObjectToInferredTypesConverter.cs`:

https://github.com/Mimetis/Dotmim.Sync/blob/e8948f7983901550b14647eedd5b2e11f04a290b/Projects/Dotmim.Sync.Core/Serialization/ObjectToInferredTypesConverter.cs#L14-L23

`JsonReader.cs`:

https://github.com/Mimetis/Dotmim.Sync/blob/e8948f7983901550b14647eedd5b2e11f04a290b/Projects/Dotmim.Sync.Core/Serialization/JsonReader.cs#L196-L208

`LocalJsonSerializer.cs` :

https://github.com/Mimetis/Dotmim.Sync/blob/e8948f7983901550b14647eedd5b2e11f04a290b/Projects/Dotmim.Sync.Core/Serialization/LocalJsonSerializer.cs#L436-L449


In your PR, you are adding the type to distinguish `DateTime` and `DateTimeOffset`.
That make sense only if we are doing the same for all types, otherwise we are lacking consistency depending on types.

I would prefer to let it as is.
**NewtonSoft.Json** is doing almost the same. 
Everything is deserialized to the closest type.
> TBS, **NewtonSoft.Json** is smart enough to manage `DateTime` and `DateTimeOffset` with the `Newtonsoft.Json.JsonSerializerSettings` settings object.

In our case, I guess `DateTimeOffset` is the way to go, don't you think ?

Just to notice that the default for `System.Text.Json.Serialize()` method is to serialize dates to string.

symbiogenesis:
ok, modified it to standardize around DateTimeOffset with a very simple change

symbiogenesis:
@Mimetis I like the performance improvements you made. Also, it centralizes these conversions for further performance work to provide broader benefits.

PR #1196, opened 2024-07-01
Correctly dispose resources

symbiogenesis:
Found via InferSharp

Ready to merge. The only test failure is fixed in another PR


Mimetis:
Sounds good to me !
Thanks for all the efforts you put in this library :)

PR #1198, opened 2024-07-01
Fix test failure reporting from Azure to GitHub

symbiogenesis:
If this works, there should be an error reported on this PR, until the tests are all green on master branch


Mimetis:
Sounds good to me !
I remember why I put the `continueOnError: true` condition, back in the days.
I wanted to make a big run and see all the errors.
But I guess now that everything is more or less stabilized, we can stop each run on the first error raised .

Good to go !

PR #1210, opened 2024-07-05
turn on lots of analyzers

symbiogenesis:
This turns on tons of static analysis, and thus adds many thousands of warnings.

To me, it is better to turn them all on, and fix them or silence them explicitly. Rather than not to know about them.

The latest StyleCop beta has the best support for C# 12, which is why I am adding it rather than relying on the built-in version.

`<NoWarn>` can be added at the Directory.Build.props level to apply everywhere for those that you will inevitably want to silence. As long as you use the inheritance pattern I showed in #1209 

Silencing those relating to documentation and naming conventions will probably take care of more than 80% of the warnings, and you can decide which to silence.


Mimetis:
I have merged the PR in main
I'll take the time next week to look at the rules enabled and see what are the ones I would like to disable

Mimetis:
God, it took me 1 month to make all the changes....
I don't know if it's worth it, but god, it's done....

PR #1234, opened 2024-08-31
Only query table list once

symbiogenesis:
This is a redundant LINQ query and list creation operation that is happening all the time.

Threw in another fix for a possible memory leak


Mimetis:
thx !

PR #1241, opened 2024-09-11
remove redundant CloseFile calls

symbiogenesis:
The `Dispose` and `DisposeAsync` methods already try to close the file because of the `using` statement.

Indeed, the `using` statement is superior, because it will close the file even in the event of an exception, and obviates the need for try-finally logic.


symbiogenesis:
Recommend retrying the one failed test

Mimetis:
Hum I like to see the `CloseFileAsync` method called, but indeed the using statement is enough.

PR #1242, opened 2024-09-11
Simplify orchestrator code

symbiogenesis:
The most complex method in the project is the `InternalApplyTableChangesAsync` method.

Anything that can be done to simplify it would be useful.


Mimetis:
Yes, the `Orchestrator` class is heavy.
I often wonder if it could be a good idea to refactor it a little bit.

Not sure it's worth it as I don't want to break something that is actually working fine

But still in my todo list 

Thanks for these little ameliorations

PR #1243, opened 2024-09-11
use structured logging, not interpolation

symbiogenesis:
No severities were changed.

see also: https://stackoverflow.com/questions/52200484/why-logging-doesnt-use-string-interpolation


Mimetis:
Make sense, let's go for it

PR #1244, opened 2024-09-11
use pattern matching

symbiogenesis:
A continuation of the simplification in https://github.com/Mimetis/Dotmim.Sync/pull/1242


Mimetis:
that's nice, I should use it more often, I guess !

PR #1245, opened 2024-09-11
Refactor ApplyRow Orchestrator

symbiogenesis:


Projects/Dotmim.Sync.Core/Orchestrators/BaseOrchestrator.cs
```diff
@@ -445,5 +445,21 @@ internal SyncException GetSyncError(SyncContext context, Exception exception, st
...
+            // If we have a transient error happening, and we are rerunning the transaction,
+            // raising an interceptor
+            var onRetry = new Func<Exception, int, TimeSpan, object, Task>((ex, cpt, ts, arg) =>
+                this.InterceptAsync(new TransientErrorOccuredArgs(context, connection, ex, cpt, ts), progress, cancellationToken).AsTask());
+
+            return this.Options.TransactionMode switch
+            {
+                TransactionMode.AllOrNothing => SyncPolicy.WaitAndRetry(0, TimeSpan.Zero),
+                TransactionMode.PerBatch => SyncPolicy.WaitAndRetryForever(retryAttempt => TimeSpan.FromMilliseconds(retryIntervalMilliseconds * retryAttempt), (ex, arg) => this.Provider.ShouldRetryOn(ex), onRetry),
+                TransactionMode.None => SyncPolicy.WaitAndRetryForever(retryAttempt => TimeSpan.FromMilliseconds(retryIntervalMilliseconds * retryAttempt), (ex, arg) => this.Provider.ShouldRetryOn(ex), onRetry),
```
symbiogenesis, line 460: @Mimetis In the original code, there was an inconsistency with the retry policy.

The `WaitAndRetryForever` logic would run when `TransactionMode` is `AllOrNothing` in two cases. And would run when **_not_** `AllOrNothing` in the other case.

Also, this means that with a transaction mode of None, this will still do a retry. But, when TransactionMode is None, that seems really destructive to keep trying.

symbiogenesis: I recommend a consistent policy and to not retry when the mode is `None`, but to just throw the exception.

Issue #1298, opened 2025-01-29
No way to detect progress for creating a snapshot


damien-c-d:
We have an app using DMS ASP.NET Core Web Proxy hosted in Azure App Service which has a non-configurable 230 second timeout for requests. We have found that some of our larger databases are hitting this timeout on initializing a new client and are in the process of configuring snapshots to be creatable before initialization. For this we will use SignalR but would like to be able to show some progress to the user. 

I can see in the code for [RemoteOrchestrator.Snapshots.cs ](https://github.com/Mimetis/Dotmim.Sync/blob/d2718d70504e3a75d30b58fc9e46614d1cdd08c4/Projects/Dotmim.Sync.Core/Orchestrators/RemoteOrchestrator.Snapshots.cs) that there is no overload that takes an `IProgress<ProgressArgs>`, however, an optional parameter could be added to the [method call](https://github.com/Mimetis/Dotmim.Sync/blob/d2718d70504e3a75d30b58fc9e46614d1cdd08c4/Projects/Dotmim.Sync.Core/Orchestrators/RemoteOrchestrator.Snapshots.cs#L62) `GetConnectionAsync` as shown below:
```csharp
using var runner = await this.GetConnectionAsync(context, SyncMode.WithTransaction, SyncStage.SnapshotCreating, progress: progress).ConfigureAwait(false);
```

The rest of the method appears to already use the IProgress that was passed to it via runner.Progress.

I have created a fork with [this change](https://github.com/damien-c-d/Dotmim.Sync/commit/82d53128152cff91794dd2ddd4ef83273f36053a) and tested in the SampleConsole app and it appears to work correctly.

Would you mind if I created a PR for this change?

Mimetis:
That sounds good to me :)

Would appreciate if you make a PR !

damien-c-d:
> That sounds good to me :)
> 
> Would appreciate if you make a PR !

Thanks! I have opened a [PR for this](https://github.com/Mimetis/Dotmim.Sync/pull/1299)

PR #1299, opened 2025-01-29
Add progress reporting for snapshot creation

damien-c-d:
This commit adds an IProgress<ProgressArgs> parameter to the CreateSnapshotAsync() methods on the RemoteOrchestrator and passes it to the GetConnectionAsync() method call.

It allows for progress reporting for creating a snapshot as mentioned in #1298


Mimetis:
thanks for the PR !

PR #1330, opened 2025-07-30
Add Exclude columns feature

wismna:
Currently, when no columns are specified in the setup, it will sync all the table columns.

This adds an column exclusion feature, so that all columns except the excluded ones will be synchronized. It allows for simpler configuration when tables have many columns and we only want to exclude a few.
The excluded columns will not be taken into account at all if columns are specified in the setup.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -15,7 +15,7 @@
 		<!-- Restore the exact packages as listed in the lock file -->
 		<RestoreLockedMode Condition="'$(ContinuousIntegrationBuild)' == 'true'">true</RestoreLockedMode>
 
-		<Version>1.3.0</Version>
+		<Version>1.3.1</Version>
 		<VersionSuffix>$(VersionSuffix)</VersionSuffix>
 		<Version Condition=" '$(VersionSuffix)' != '' ">$(Version)-$(VersionSuffix)</Version>
 
--- a/Projects/Dotmim.Sync.Core/Orchestrators/BaseOrchestrator.Schema.cs
+++ b/Projects/Dotmim.Sync.Core/Orchestrators/BaseOrchestrator.Schema.cs
@@ -247,13 +247,11 @@ namespace Dotmim.Sync
                 if (schemaTable.Columns.Count > 0)
                     schemaTable.Columns.Clear();
 
-                List<SyncColumn> lstColumns;
+                List<SyncColumn> lstColumns = new List<SyncColumn>();
 
                 // Validate columns list from setup table if any
                 if (setupTable.Columns != null && setupTable.Columns.Count > 1)
                 {
-                    lstColumns = new List<SyncColumn>();
-
                     foreach (var setupColumn in setupTable.Columns)
                     {
                         // Check if the columns list contains the column name we specified in the setup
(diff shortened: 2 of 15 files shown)


wismna:
> Thanks for the contribution ! Appreciate !

You're welcome !

> 1. We need tests :)

Indeed we do, I'll get on it.

> 2. What happens if we have this code:
> 
> ```cs
> var setup = new SyncSetup("Customer");
> 
> // Add columns
> setup.Tables["Employee"].Columns.AddRange("CustomerID", "FirstName", "LastName", "Title", "Suffix", "CompanyName", "rowGuid");
> 
> setup.Tables["Employee"].Columns.ExcludeRange("rowGuid", "LastModifiedDate");
> ```
> 
> I know this example is kind of stupid, but if you think about some algorithm where colums are built using a loop or any dynamic routine, we can end up with such situation

Well, I asked myself that same question, and I'm not sure there is a "right" answer. The way I see it, we have three possibilities:
- included columns win
- excluded columns win
- we raise an error, which is what I tried to implement (cf. last remark)

What's your opinion on the subject ?

> 3. For adding columns, we have 2 main method : `SetupColumns.Add(string item)` and `SetupColumns.AddRange(params string[] columnsName)`
> 
> If you look carefully `SetupColumns.Add(string item)` method (called from `AddRange`) is here to ensure columns name is parsed to avoid incorrect character and then ensure column exists. Even if we are excluding a column, we need to ensure that these two verifications are done. Indeed if someone mispells a column name to be excluded, the column will not be excluded eventually...

Oh, right, I missed that ! I'll fix it.

> By the way in method `Add()`, you are checking `ExcludedCollection`, and that makes no sense to me !

Perhaps I did not do that properly, but it's related to your second point: if we try to add a column that is already excluded, we raise an error as it kind of made no sense for me to have the same column both included and excluded.

Mimetis:
> > I know this example is kind of stupid, but if you think about some algorithm where colums are built using a loop or any dynamic routine, we can end up with such situation
> 
> Well, I asked myself that same question, and I'm not sure there is a "right" answer. The way I see it, we have three possibilities:
> 
> * included columns win
> * excluded columns win
> * we raise an error, which is what I tried to implement (cf. last remark)
> 
> What's your opinion on the subject ?

I guess we can use a kind of merge method

- If you Include columns, then exclude columns, excluding is the last action and is the winner :
Example: Include C1 C2 C3 C4 then Exclude C2 => C1 C3 C4 is the final result

- if you exclude columns then include, including is the last action and is the winner
Example: Exclude C1 C2 then Include C2 C3 C4 C5 => C2 C3 C4 C5 is the winner

By the way, raising an error can also be a solution ... I don't know !

> > By the way in method `Add()`, you are checking `ExcludedCollection`, and that makes no sense to me !
> 
> Perhaps I did not do that properly, but it's related to your second point: if we try to add a column that is already excluded, we raise an error as it kind of made no sense for me to have the same column both included and excluded.

True !

wismna:
Hi !

So, I updated the code. After our discussions, I figured that I would implement it in a way where the last operation wins, as you described here:

> if you Include columns, then exclude columns, excluding is the last action and is the winner :
Example: Include C1 C2 C3 C4 then Exclude C2 => C1 C3 C4 is the final result
>
> if you exclude columns then include, including is the last action and is the winner
Example: Exclude C1 C2 then Include C2 C3 C4 C5 => C2 C3 C4 C5 is the winner

Also, I added unit tests to cover these cases and used the same code to parse column name.

Let me know what you think !

PR #1374, opened 2025-12-21
Modify WHERE clause in CreateUpdateCommandText to handle updates with no tracking rows

### LINKED ISSUE #1373: SqliteSyncProvider always reports a conflict when applying changes where no tracking row exists
As per this discussion: https://github.com/Mimetis/Dotmim.Sync/discussions/1372, it was discovered that the SqliteSyncProvider will always report conflicts when:

* A row exists in a client SQLite DB
* The row is updated elsewhere and synced to the client
* There is no _tracking entry for the row

With much digging and AI assistance the faulty code was identified:

https://github.com/Mimetis/Dotmim.Sync/blob/f4378baac80c262d820976e2f0fc3ba5fd70a487/Projects/Dotmim.Sync.Sqlite/Builders/SQLiteObjectNames.cs#L358-L361

This creates a WHERE clause:

``` sql
LEFT JOIN [Row_tracking] AS [side] ON [side].[Id] = @Id
LEFT JOIN [Row] AS [base] ON [base].[Id] = @Id
WHERE ([side].[timestamp] <= @sync_min_timestamp OR [side].[update_scope_id] = @sync_scope_id) 
OR ([base].[Id] IS NULL  AND ([side].[timestamp] <= @sync_min_timestamp OR [side].[timestamp] IS NULL)) 
```

Which has the following issue:

```
Scenario A: The Row DOES NOT exist (New Record)
If [base].[Id] is NULL (it's a new insert):

[base].[Id] IS NULL is TRUE.
Inside the inner parentheses:
[side].[timestamp] <= ... is FALSE.
[side].[timestamp] IS NULL is TRUE (Because there is no tracking row).
The logic becomes: TRUE AND (FALSE OR TRUE) -> TRUE.
Outcome: The row IS selected. The sync proceeds to insert the new row.

Scenario B: The Row ALREADY exists
If [base].[Id] is NOT NULL (the row is there):

[base].[Id] IS NULL is FALSE.
The AND condition immediately fails.
Outcome: The row IS NOT selected. The sync skips this update (unless the @sync_force_write flag from the wider query was set to 1).
```

This is resolved by modifying the SQL builder to handle cases correctly:

``` cs 
// 1. Standard Update: Only allow if row exists (NOT NULL) AND Tracking conditions are met
// We verify [base] is NOT NULL to ensure we don't implicitly resurrect a tombstoned row here.
stringBuilder.Append($"WHERE (NOT ({SqliteManagementUtils.WhereColumnIsNull(this.TableDescription.PrimaryKeys, "[base]")}) ");
stringBuilder.AppendLine($"AND ([side].[timestamp] <= @sync_min_timestamp OR [side].[update_scope_id] = @sync_scope_id)) ");

// 2. Missing Tracking: Allow write if Tracking row is missing (New Inserts or Data Update)
stringBuilder.AppendLine($"OR ({SqliteManagementUtils.WhereColumnIsNull(this.TableDescription.PrimaryKeys, "[side]")}) ");

// 3. Force Write
stringBuilder.Append($"OR @sync_force_write = 1");
```

kmgallahan:
These changes modify `CreateUpdateCommandText` to properly handle updates where a row exists but it has no _tracking entry.

Resolves #1373

### DIFF
--- a/Projects/Dotmim.Sync.Sqlite/Builders/SQLiteObjectNames.cs
+++ b/Projects/Dotmim.Sync.Sqlite/Builders/SQLiteObjectNames.cs
@@ -355,9 +355,9 @@ namespace Dotmim.Sync.Sqlite
             stringBuilder.AppendLine($"FROM (SELECT {stringBuilderParametersValues}) as [c]");
             stringBuilder.AppendLine($"LEFT JOIN {this.TrackingTableQuotedShortName} AS [side] ON {str1}");
             stringBuilder.AppendLine($"LEFT JOIN {this.TableQuotedShortName} AS [base] ON {str2}");
-            stringBuilder.AppendLine($"WHERE ([side].[timestamp] < @sync_min_timestamp OR [side].[update_scope_id] = @sync_scope_id) ");
-            stringBuilder.Append($"OR ({SqliteManagementUtils.WhereColumnIsNull(this.TableDescription.PrimaryKeys, "[base]")} ");
-            stringBuilder.AppendLine($"AND ([side].[timestamp] < @sync_min_timestamp OR [side].[timestamp] IS NULL)) ");
+            stringBuilder.Append($"WHERE (NOT ({SqliteManagementUtils.WhereColumnIsNull(this.TableDescription.PrimaryKeys, "[base]")}) ");
+            stringBuilder.AppendLine($"AND ([side].[timestamp] <= @sync_min_timestamp OR [side].[update_scope_id] = @sync_scope_id)) ");
+            stringBuilder.AppendLine($"OR ({SqliteManagementUtils.WhereColumnIsNull(this.TableDescription.PrimaryKeys, "[side]")}) ");
             stringBuilder.Append($"OR @sync_force_write = 1");
             stringBuilder.AppendLine($")");
 


Mimetis:
I think I found something easier to understand

``` cs
            stringBuilder.AppendLine($"WITH CHANGESET as (SELECT {stringBuilderParameters} ");
            stringBuilder.AppendLine($"FROM (SELECT {stringBuilderParametersValues}) as [c]");
            stringBuilder.AppendLine($"LEFT JOIN {this.TrackingTableQuotedShortName} AS [side] ON {str1}");
            stringBuilder.AppendLine($"WHERE ([side].[timestamp] < @sync_min_timestamp OR [side].[update_scope_id] = @sync_scope_id OR [side].[timestamp] IS NULL) ");
            stringBuilder.Append($"OR @sync_force_write = 1");
            stringBuilder.AppendLine($")");
```

This version remove the join on the table itself (`$"LEFT JOIN {this.TableQuotedShortName} AS [base] ON {str2}"`)

We just need to know:
- If we are below the timestamp
- Or we are the author of the row
- Or we don't have row

Then the Insert / Conflict update will take care of the state of the row (existing or not)

kmgallahan:
I think we need to check if the data row exists because as mentioned:

```
// We verify [base] is NOT NULL to ensure we don't implicitly resurrect a tombstoned row here.
```

When a client deletes a data row a _tracking row representing the tombstoned data row is created, while at the same time the data row gets deleted. According to `BaseOrchestrator.ApplyChanges` - Inserts and Updates and processed BEFORE Deletes. This means that if we ignore existence of the data row and don't check the _tracking row's tombstoned flag then the data row can get re-inserted into the client database as if it were an INSERT.

Also note that there is a discrepancy in the timestamp comparison logic: SQL Server provider uses <=, while currently the SQL Lite provider only uses <. I updated the Lite provider to match the comparison logic, although this may not be correct.

Mimetis:
Sounds good
I work on the test and let you know.
I guess we still can look at the base row "just in case"

Will update the main branch before end of year

kmgallahan:
Edited:

We could do this to check the sync_row_is_tombstone flag instead of using the extra table join - it should accomplish the same thing:

``` cs
stringBuilder.AppendLine($"WITH CHANGESET as (SELECT {stringBuilderParameters} ");
stringBuilder.AppendLine($"FROM (SELECT {stringBuilderParametersValues}) as [c]");
stringBuilder.AppendLine($"LEFT JOIN {this.TrackingTableQuotedShortName} AS [side] ON {str1}");
stringBuilder.AppendLine($"WHERE ([side].[sync_row_is_tombstone] IS NULL OR [side].[sync_row_is_tombstone] <> 1)");
stringBuilder.AppendLine($"AND (([side].[timestamp] <= @sync_min_timestamp OR [side].[update_scope_id] = @sync_scope_id OR [side].[timestamp] IS NULL) ");
stringBuilder.AppendLine($"OR @sync_force_write = 1))");
```

to generate this:

``` sql
WITH CHANGESET as (SELECT ... 
FROM (SELECT ...) as [c]
LEFT JOIN [TrackingTable] AS [side] ON ...
WHERE ([side].[sync_row_is_tombstone] IS NULL OR [side].[sync_row_is_tombstone] <> 1)
AND (([side].[timestamp] <= @sync_min_timestamp OR [side].[update_scope_id] = @sync_scope_id OR [side].[timestamp] IS NULL) 
OR @sync_force_write = 1))
```

*I'm unsure if `sync_force_write` should be able to resurrect deleted rows.

